PR #15, opened 2015-11-03
Add functions to set or get a Random Forest model

elehcim:
The type ForestInfo has been introduced as a vector of SplitMatrix.

The functions `getForestInfo()` and `setForestInfo` were added.

The constructor of `CARTClassifier` has been modified so that member `m_splitMatrix` is filled when creating a CARTClassifier [here](https://github.com/Shark-ML/Shark/blob/c181701a5882ac85c5be2354bf57e12c2f4b54a7/src/Algorithms/RFTrainer.cpp#L135).

For convenience, I also added a verbosity flag to print the status of the training process (this part of the pull request can be ignored)

### DIFF
--- a/include/shark/Models/Trees/CARTClassifier.h
+++ b/include/shark/Models/Trees/CARTClassifier.h
@@ -100,7 +100,16 @@ public:
 	/// Constructor taking the splitMatrix as argument
 	CARTClassifier(SplitMatrixType const& splitMatrix)
 	{
-		setSplitMatrix(splitMatrix);
+		m_splitMatrix=splitMatrix;
+	}
+
+	/// Constructor taking the splitMatrix as argument and optimize it if requested
+	CARTClassifier(SplitMatrixType const& splitMatrix, bool optimize)
+	{
+		if (optimize)
+			setSplitMatrix(splitMatrix);
+		else
+			m_splitMatrix=splitMatrix;
 	}
 
 	/// Constructor taking the splitMatrix as argument as well as maximum number of attributes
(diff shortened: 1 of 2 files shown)


include/shark/Models/Trees/CARTClassifier.h
```diff
@@ -100,7 +100,7 @@ class CARTClassifier : public AbstractModel<RealVector,LabelType>
 	/// Constructor taking the splitMatrix as argument
 	CARTClassifier(SplitMatrixType const& splitMatrix)
 	{
-		setSplitMatrix(splitMatrix);
+		m_splitMatrix=splitMatrix;
```
Ulfgard: Is this change here on purpose avoiding the call to optimizeSplitMatrix by setSplitMatrix? What would be the rationale behind that?

Ulfgard:
thanks for the pull request. I hope the amount of comments was okay. I have opened issue #16 to reflect the issue with the print out

include/shark/Models/Trees/CARTClassifier.h
elehcim: Yes, that's because when loading a model that was exported using this code, the splitMatrix is already optimized, and optimizeSplitMatrix was not needed. I experienced a big bottleneck when loading a model using setSplitMatrix. (Sorry, I did this long time ago, I added a misleading line in the PR description).
But I understand, to be safer it's better to stay with `setSplitMatrix(splitMatrix)`.
Since I use it [here](https://github.com/elehcim/Shark/commit/24b0a186682e83594ccd1226fa70188a945eef03#diff-6811b3d4bf9cbf971149c0ba3342dd73R137), maybe I can add an `operator=` without optimizeSplitMatrix and leave this constructor as is.

Ulfgard: I see. Maybe we can add a second argument "bool optimize" (i think better without default argument) instead. Then the user decides whether he wants to optimize or not.

elehcim:
Request for comments. When done I can squash the commits

Ulfgard:
hi, sorry for the late reply! this somehow went under my radar. Looks good. could you squash it?

elehcim:
Thanks. It's squashed, I rebased it on top of v3.0.1.

Ulfgard:
thank you!

PR #18, opened 2015-11-04
Openblas

skn123:
Check out the ReadMe for building OpenBlas on windows! Once that is done, the task becomes easy.
I have third party libraries in my build environment. You can uncheck "SHARK_CUSTOMIZED_LIBS" and check "SHARK_DEFAULT_LIBS" and continue working as is. I am using Boost 1.55 and now moving to 1.59 ! My hdf5 is also the latest from the svn repo.


Ulfgard:
Hi, thanks for pull request

this whole cblas thing seems to be a mess. Can I conclude that the only changes you needed to include were those that changed types between ATLAS<->CBLAS as well as some LACPACK/LAPACKE things?

my linux openblas package (As well as atlas) relies on the standard interface as defined here:
http://www.netlib.org/blas/blast-forum/cinterface.pdf

so your changes can not run on my laptop :(

I need to find a way to fix this and maybe this is going to be a lot easier than basically copy pasting the whole thing.

If I do not come back to you tomorrow, I will be on vacation over the next week.

skn123:
Yes, that is all that is needed. Barring that all it needs is OpenBlas to be built/found and linked with Shark. I can run all the examples without any problem

PR #23, opened 2015-11-23
Fix HDF5 detection.

ghisvail:
This PR fixes HDF5 detection on Linux and probably other platforms. According to the upstream CMake documentation for [FindHDF5](https://cmake.org/cmake/help/v3.0/module/FindHDF5.html), the module does not define a `HDF5_HL_COMPILER_EXECUTABLE` variable. The latter is thus removed from the tests.


Ulfgard:
Thank you. sorry for my absense for a few days, I am now going through your reports

ghisvail:
Did you have a chance to look at this PR ?

PR #48, opened 2016-02-23
Usage of complex matrices with Cholesky decomposition

didacr:
I had to change the kernel function because the arguments were not matching the proper call, and the potrf function to the proper complex logic. It does not look pretty, though.


Ulfgard:
Hi Didac,

the bindings part looks good. The default implementation does not, because it would return false for all complex numbers. the right logic is: if the real part is positive and the complex part 0, than we can use sqrt.

Be aware that you can still not solve systems of equations, though as you need the triangular solver, especially with the conjugate transpose of the factor.

PR #49, opened 2016-02-23
Interface evo

Ulfgard:
We reworked the EA interfaces and split the step into different phases: generating offspring and updating parameters. This way a derived class can get deeper insight into the working of the algorithm.

we also fixed a lot of algorithms when we were at it.

### DIFF
--- a/examples/EA/SOO/CMASimple.tpp
+++ b/examples/EA/SOO/CMASimple.tpp
@@ -52,8 +52,8 @@ int main( int argc, char ** argv ) {
 	// Initialize the optimizer for the objective function instance.
 	//###begin<optimizer>
 	CMA cma;
-	cma.setInitialSigma(0.1);
-	cma.init( sphere, sphere.proposeStartingPoint()); // Explicitely set initial global step size.
+	cma.setInitialSigma(0.1);// Explicitely set initial global step size.
+	cma.init( sphere, sphere.proposeStartingPoint()); 
 	//###end<optimizer>
 
 	// Iterate the optimizer until a solution of sufficient quality is found.
--- a/examples/EA/SOO/CMASimple.tpp
+++ b/examples/EA/SOO/CMASimple.tpp
@@ -52,7 +52,8 @@ int main( int argc, char ** argv ) {
 	// Initialize the optimizer for the objective function instance.
 	//###begin<optimizer>
 	CMA cma;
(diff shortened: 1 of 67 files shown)


examples/EA/SOO/CMASimple.tpp
```diff
@@ -52,8 +52,8 @@ int main( int argc, char ** argv ) {
 	// Initialize the optimizer for the objective function instance.
 	//###begin<optimizer>
 	CMA cma;
-	cma.init( sphere );
-	cma.setSigma( 0.1 ); // Explicitely set initial globael step size.
+	cma.setInitialSigma(0.1);
+	cma.init( sphere, sphere.proposeStartingPoint()); // Explicitely set initial global step size.
```
TGlas: It is a detail, but the comment is copy&paste from two lines up, and this is not what's happending here.

src/Algorithms/DirectSearch/CMA.cpp
```diff
@@ -160,35 +163,62 @@ void CMA::write( OutArchive & archive ) const {
...
+	}
+}
+void CMA::doInit( 
+	std::vector<SearchPointType> const& initialSearchPoints,
+	std::vector<ResultType> const& initialValues,
+	std::size_t lambda,
+	std::size_t mu,
+	double initialSigma
+) {
+	m_numberOfVariables =initialSearchPoints[0].size();
```
TGlas: Please assert if initialSearchPoints is empty

include/shark/Algorithms/DirectSearch/CMSA.h
```diff
@@ -51,118 +51,131 @@
...
+	
+	/// \brief sets the initial step length sigma
+	///
+	/// It is by default <=0 which means that sigma =1/sqrt(numVariables)
+	void setInitialSigma(double initSigma){
+		m_initSigma = initSigma;
+	}
+
+	/// \brief Sets the number of selected samples
+	void setMu(std::size_t mu){
```
TGlas: Does it make sense to provide a method?
I am not sure, this may be unnecessary interface bloat:
void setMuToDefault() { m_userSetMu = false; }

Ulfgard: yes. it does make sense. the provided init method taking all the hyper parameters can not be provided as a virtual interface. so we need a way to set some parameters before init() is called(e.g. the user creates the optimizer, sets the hyper parameters and then provides it to the remaining program)

PR #65, opened 2016-03-30
Non dominated sorting

TGlas:
* simplified the Pareto dominance relation
* implemented an alternative non-dominated sorting algorithm, suitable for large populations and for low dimensions (few objectives)
* added an abstraction for switching between non-dominated sorting algorithms
* improved the non-dominated sorting unit test


Ulfgard:
A unit test for Hypervolume is failing. Do you know, why? maybe it uses some pareto domiannce relation?

I would suggest moving the actual NonDom sort algorithms in a subfolder (as we did with operators/Hypervolume/). Further you can make those independent from the extractor by using boost::adaptors::transform:

http://www.boost.org/doc/libs/1_60_0/libs/range/doc/html/range/reference/adaptors/reference/transformed.html

The code would then read for example:

```
FastNonDominatedSort sorter;
sorter(boost::adaptors::transform(pop,e));
```

I could help with that

chivesman:
Looks great in general. A minor thing:

`BOOST_CHECK(population[i].rank() < population[j].rank() )`

Might be better to write:
`BOOST_CHECK_LT(population[i].rank(), population[j].rank() )`

So that the two numbers will be reported in case of failure.

TGlas:
Thanks for the comments. I'll change the `BOOST_CHECK` macro, and I'll move the files into a sub-folder.

The failing unit test is annoying. Last time I checked everything still worked (iirc), but then I kept running only DirectSearch_NonDominatedSort during development. It seems that in all failing cases the hypervolume is infinite...? Weird, I'll investigate.

I am not so sure about the transformation approach. Writing this out is cumbersome. Compare
```
FastNonDominatedSort sorter;
FitnessExtractor e;
sorter(boost::adaptors::transform(pop,e));
```
to
```
FastNonDominatedSort<FitnessExtractor> sorter;
sorter(pop);
```
I rate usability (not recalling and typing the boost stuff) higher than getting rid of a tempate parameter.

Actually, I did not change the interfaces, but I don't really like them. Extractors don't have a state, therefore they should be free (template) functions. The same holds for the sorters. Then one could do:
`nonDominatedSort(pop, fitnessExtractor); // no need to create sorter and extractor objects`
possibly with a default argument for the extractor. We can try to model the extractor as a std::function. Comments and opinions welcome!

Ulfgard:
Its not about a template parameter.

The idea of the transformation approach is that the implementation of the algorithm becomes easier. Instead of having to work with the extractor, we an just pretend, the argument is "somehow equivalent to std::vector<RealVector>". The frontend class NonDominatedSort(?) can then still use an extractor and just transform the population with it.

I am honest, I do not like the extractor interface. It starts with that you have to have an "identityFitnessExtractor" if you just want to use it with a simple set of points. And it is everywhere.

The range approach instead only adds complexity where the transformation is needed. Of course, writing the line out like this is a burden, but why not have a penalizedFitness(pop) or searchPoint(pop) function? its literally 6 lines of code and covers 100% of use-cases where the extractor is not the identity.

So my goal is to fade this out, starting with the points that actually call the extractor.

TGlas:
Hi,

> Its not about a template parameter.
>
> The idea of the transformation approach is that the implementation of the algorithm becomes easier. Instead of having to work with the extractor, we an just pretend, the argument is "somehow equivalent to std::vector<RealVector>". The frontend class NonDominatedSort(?) can then still use an extractor and just transform the population with it.

I see. Seems to make sense. For possible caveats see below.

> I am honest, I do not like the extractor interface. It starts with that you have to have an "identityFitnessExtractor" if you just want to use it with a simple set of points. And it is everywhere.

I don't think that IdentityFitnessExtractor is of any use other than
in unit tests. In Shark 2 there was a bool parameter indicating whether
penalized or unpenalized fitness should be used. This was leading to
if-then-else everywhere down the line, which was much worse than the
extractor. So yes, I see the point, but it could be much worse :)

> The range approach instead only adds complexity where the transformation is needed. Of course, writing the line out like this is a burden, but why not have a penalizedFitness(pop) or searchPoint(pop) function? its literally 6 lines of code and covers 100% of use-cases where the extractor is not the identity.

At some point I have to write the result back, which right now works
with a call like
    pop[i].rank() = r;
Now if pop[i] is a RealVector then we have to restructure far more
than it seems at first glance. For example, we could return a vector
of ranks which then must be assigned to the population from outside
the non-dominated sorting function. Then the extractor sounds like the
better approach.

> So my goal is to fade this out, starting with the points that actually call the extractor.

I am open to this suggestion, but I yet understand know how this would
work.

Tobias

Ulfgard:
Hi,

Thinking about it, I also do not like that the concept of pareto fronts is tied to individuals. Maybe we should change this as well. The easiest solution would be to take the range of ranks as a second argument.

Then calling it could be something like:

sorter(penalizedFitness(pop), ranks(pop));

TGlas:
Hi,

> Thinking about it, I also do not like that the concept of pareto fronts is tied to individuals. Maybe we should change this as well. The easiest solution would be to take the range of ranks as a second argument.
>
> Then calling it could be something like:
>
> sorter(penalizedFitness(pop), ranks(pop));

Okay, that solution looks good to me. Indeed, very generic.
I would vote for nonDominatedSort(penalizedFitness(pop), ranks(pops));
i.e., without the sorter object.

What is the ranks(pop) object? It should be a vector<std::size_t&>,
but that's not going to work. So we need a proxy, right?

Tobias

Ulfgard:
rank(pop) should be a shortcut for (out of my head, so beware of errors)

boost::adaptors::transform(pop,[](Individual& ind) ->int& {return ind->rank();});

Ulfgard:
if you give it a go, i could do this

TGlas:
> if you give it a go, i could do this
Go, go, go! :)

Ulfgard:
Hi,

i have just pushed my changes. unfortunately they collided with yours. I kept mine for now because i did not want to mess up my other changes. we can move the files later on if we wish.

i have done my thing. unfortunately, as with many boost libraries written pre c++11, i stumbled over some issues with boost::range... everything works, but it will not work with lambdas for now.

Ulfgard:
if you are happy with the state just merge :) i fixed the failing test and my raised concerns are resolved.

PR #69, opened 2016-04-09
Removed Extractor Interface

Ulfgard:
....and cleaned up indicators a little bit.

This pull request removes the extractor interface. This has the following changes:

-Selection: instead of an extractor an ordering relation is used. There is not the assumption an more that there is a "rank" or "fitness" to be querried.

- Indicators now expect a range of fitness vectors and return the index of the vector which is the least contributor. (I also tried to clean up the least contributor approximator which is not in working condition any way, so it probably does not work...i will change that soon)

- In diverse files there where still extractors, e.g. the cog() functions used and re-implemented everywhere...


examples/EA/SOO/TSP.tpp
```diff
@@ -81,149 +81,144 @@ const double cities[10][10] = {
...
+	RealMatrix m_costMatrix;
+};
+
+/**
+ * @brief Implements partially mapped crossover
+ * 
+ * Makes sure that only correct tours on the graph
+ * are generated.
+ */
+struct PartiallyMappedCrossover {
```
TGlas: I did not even know that we have an implementation of PMX. This should move into its own file in Operators/Recombination/PMX.h

examples/EA/SOO/TSP.tpp
```diff
@@ -243,21 +238,21 @@ int main( int argc, char ** argv ) {
 	}
 
 	// Mating selection operator
-	shark::RouletteWheelSelection rws;
+	RouletteWheelSelection rws;
 	// Variation (crossover) operator
-	shark::PartiallyMappedCrossover pmc;    
+	PartiallyMappedCrossover pmc;
```
TGlas: I would really rename this variable into pmx, since that's the usual abbreviation of partially mapped crossover.

include/shark/Algorithms/DirectSearch/Indicators/HypervolumeIndicator.h
```diff
@@ -29,128 +29,70 @@
...
-	*/
-	template<typename Extractor, typename ParetoFrontType, typename VectorType>
-	std::size_t leastContributor( Extractor const& extractor, ParetoFrontType const& front, VectorType const& referencePoint)
+	/// \brief Determines the point contributing the least hypervolume to the overall front of points.
+	///
+	/// This version uses the reference point estimated by the last call to updateInternals.
+	///
+	/// \param [in] front pareto front of points
+	template<typename ParetoFrontType>
+	std::size_t leastContributor( ParetoFrontType const& front)
```
TGlas: Would it make sense here to return an iterator instead?

include/shark/Algorithms/DirectSearch/Indicators/HypervolumeIndicator.h
```diff
@@ -29,128 +29,70 @@
...
+		std::size_t noObjectives = front[0].size();
 		m_reference.resize(noObjectives);
-		m_reference.clear();
 		
-		for( unsigned int i = 0; i < set.size(); i++ )
-			noalias(m_reference) = max(m_reference, extractor(set[i]));
+		for(auto const& point: front)
+			noalias(m_reference) = max(m_reference, point);
 		
 		noalias(m_reference) += 1.0;
```
TGlas: I still think that 1 is too much of a magic value. There needs to be some value, but the user should be responsible for setting it explicitly. This is like the C = 1 default for an SVM. Really, some parameters need a design choice and cannot have a default. If at all, then the default should be relative to the range of observed values, e.g., param * (max - min), with say, param=1.

include/shark/Algorithms/DirectSearch/Indicators/LeastContributorApproximator.h
```diff
@@ -393,94 +224,31 @@ struct LeastContributorApproximator {
 
 	}
 
-	/**
```
TGlas: Okay, you I will ignore that this is commented out since you plan to look at this soon anyway.

include/shark/Algorithms/DirectSearch/Indicators/LeastContributorApproximator.h
```diff
@@ -562,40 +330,196 @@ struct LeastContributorApproximator {
...
+			std::sort( 
+				it->m_influencingPoints.begin(),
+				it->m_influencingPoints.end(),
+				[]( SetIterator i1, SetIterator i2 ){
+					return i1->m_overlappingVolume > i2->m_overlappingVolume;
+				}
+			);
+		}
+	}
 };
```
TGlas: I did not check the large block above this point because you plan to look at this anyway soon.

include/shark/Algorithms/DirectSearch/Indicators/MultiplicativeEpsilonIndicator.h
```diff
@@ -30,87 +30,46 @@
...
-		}
-		
-		return std::min_element( relativeApproximation.begin(), relativeApproximation.end() ) - relativeApproximation.begin();
+	/// \brief Given a pareto front, returns the index of the point which is the least contributer
+	template<typename ParetofrontType>
+	unsigned int leastContributor(ParetofrontType const& front){
+		typedef decltype(points[0]) Point;
+		auto logTransform = [](Point const& x){return log(x);};
+		AdditiveEpsilonIndicator indicator;
+		return indicator.leastContributor(boost::adaptors::transform(points,logTransform));
```
TGlas: Yes, very good solution! However, shouldn't we apply exp(...) in the very end?

include/shark/Algorithms/DirectSearch/Operators/Selection/ElitistSelection.h
```diff
@@ -77,28 +77,31 @@ struct ElitistSelection {
...
 		std::size_t size = std::distance( it, itE );
-		Extractor e;
-		std::vector<KeyValuePair<double, InIterator> > individuals(size);
-		for(std::size_t i = 0; i != size; ++i){
-			individuals[i].key = e(*(it+i));
-			individuals[i].value = it+i;
-		}
-		std::sort( individuals.begin(), individuals.end());
+		std::vector<InIterator > individuals(size);
+		std::iota(individuals.begin(),individuals.end(),it);
```
TGlas: wow :)

src/Algorithms/DirectSearch/CMA.cpp
```diff
@@ -280,11 +256,16 @@ std::vector<CMA::IndividualType> CMA::generateOffspring( ) const{
...
+	ElitistSelection<IndividualType::FitnessOrdering > selection;
 	selection(offspring.begin(),offspring.end(),selectedOffspring.begin(), selectedOffspring.end());
 	m_counter++;
-	RealVector z = weightedSum( selectedOffspring, m_weights, StepExtractor() ); // eq. (38)
-	RealVector m = weightedSum( selectedOffspring, m_weights, PointExtractor() ); // eq. (39) 
+	
+	RealVector z( m_numberOfVariables, 0. );
+	RealVector m( m_numberOfVariables, 0. );
+	for( std::size_t j = 0; j < selectedOffspring.size(); j++ ){
+		noalias(z) += m_weights( j ) * selectedOffspring[j].chromosome();
```
TGlas: please preserve the equation number comments: eq. (38) and eq. (39)

TGlas:
Looks good overall, I added a few comments.

include/shark/Algorithms/DirectSearch/Indicators/HypervolumeIndicator.h
Ulfgard: i would like to change it differently at some point. Changing this was not the goal of this pull request.

we should open a ticket, discuss this and then solve it. I will not replace one ad-hoc solution by another one.

include/shark/Algorithms/DirectSearch/Indicators/MultiplicativeEpsilonIndicator.h
Ulfgard: we only return the index, not the value, so no exp needed

examples/EA/SOO/TSP.tpp
Ulfgard: can you confirm that the implementation is correct? or hint to a publication?

include/shark/Algorithms/DirectSearch/Indicators/HypervolumeIndicator.h
Ulfgard: unfortunately, that is not very helpful. right now the assumption is that the input is some kind of range of fitness values. this is a different range than the range of individuals and thus an iterator would be used to get the index of it

examples/EA/SOO/TSP.tpp
TGlas: >> +/**
>> + * @brief Implements partially mapped crossover
>> + *
>> + * Makes sure that only correct tours on the graph
>> + * are generated.
>> + */
>> +struct PartiallyMappedCrossover {
>
> can you confirm that the implementation is correct? or hint to a publication?

Here is a "reference" (extracted from the old edition of the Eiben and 
Smith book, I think):
http://www.ini.rub.de/PEOPLE/glasmtbl/lectures/EA-2015-16/lecture-3-GA-handout-1.pdf

The implemementation looks a bit weird to me - may still be correct.
The initial sampling of the crossover points could be more efficient:

SHARK_ASSERT(t1.size() == t2.size());
std::size_t cuttingPoint1 = Rng::discrete(0, t1.size() - 2);
std::size_t cuttingPoint2 = Rng::discrete(0, t1.size() - 1);
if (cuttingPoint1 >= cuttingPoint2) {
   cuttingPoint1++;
   swap(cuttingPoint1, cuttingPoint2);
}

The line with std::numeric_limits<std::size_t>::max() looks fishy,
where is this value actually assigned?

Ulfgard: done

examples/EA/SOO/TSP.tpp
Ulfgard: done

src/Algorithms/DirectSearch/CMA.cpp
Ulfgard: done

Issue #70, opened 2016-04-15
Sparse data import/export and regression labels


TGlas:
Currently sparse data ("libsvm" format) support is restricted to classification problems. This is for no apparent reason, since libsvm itself supports regression (of a single target variable). Therefore the sparse data routines should be extended to the regression case.

Ulfgard:
fixed

PR #71, opened 2016-04-17
Sparse data regression

TGlas:
This should fix the following issue
https://github.com/Shark-ML/Shark/issues/70
which I opened two days ago.


include/shark/Data/SparseData.h
```diff
@@ -116,36 +155,37 @@ SHARK_EXPORT_SYMBOL void importSparseData(
...
-	}
 
 	size_t dim = inputDimension(dataset);
 	if(numberOfClasses(dataset)!=2) oneMinusOne = false;
 
 	std::vector<detail::LabelSortPair> L;
-	if(sortLabels) {
+	if (sortLabels) {
 		for(std::size_t i = 0; i < elements; i++) 
 			L.push_back(detail::LabelSortPair(dataset.element(i).label, i));
```
Ulfgard: When we are at it, LabelSortPair can be replaced by a shark::KeyValuePair<unsigned int,double> found in Core/utility/KeyValuePair.h and instead of the constructor one can use makeKeyValuePair(dataset.element(i).label, i)

include/shark/Data/SparseData.h
```diff
@@ -157,20 +197,109 @@ void exportSparseData(LabeledData<InputType, unsigned int>& dataset, const std::
...
+		}
+		stream << std::endl;
+	}
+}
+
+/// \brief Export classification data to sparse data (libSVM) format.
+///
+/// \param  dataset     Container storing the data
+/// \param  fn          Output file name
+/// \param  dense       Flag for writing all components, including zeros
```
Ulfgard: why would i want to write a dense dataset anyways? :/

src/Data/SparseData.cpp
```diff
@@ -101,6 +98,8 @@ shark::LabeledData<T, unsigned int> libsvm_importer(
 		int maxPositiveLabel = -1;
 		for(std::size_t i = 0; i != numPoints; ++i){
 			int label = contents[i].first;
```
Ulfgard: int label = static_cast<int>(contents[i].first); o be consistent with below

include/shark/Data/SparseData.h
TGlas: > why would i want to write a dense dataset anyways? :/

VERY good question. I just tried to keep compatibility with the existing 
interface, but my first reflex was: remove!

src/Data/SparseData.cpp
```diff
@@ -37,43 +37,40 @@
...
 
-typedef std::pair<int, std::vector<std::pair<std::size_t, double> > > LibSVMPoint;
+typedef std::pair<double, std::vector<std::pair<std::size_t, double> > > LibSVMPoint;
 inline std::vector<LibSVMPoint> 
-importSparseDataReader(
-	std::istream& stream
-) {	
-	std::vector<LibSVMPoint>  fileContents;
-	while(stream){
+importSparseDataReader(std::istream& stream) {
```
Ulfgard: for VC++ it might help to put this into a separate compilation unit and add a header entry for it. Spirit takes a lot of ressources during compilation

src/Data/SparseData.cpp
```diff
@@ -138,7 +137,63 @@ shark::LabeledData<T, unsigned int> libsvm_importer(
...
+	std::size_t numPoints = contents.size();
+	
+	//find data dimension by getting the maximum index
+	std::size_t maxIndex = 0;
+	for(std::size_t i = 0; i != numPoints; ++i){
+		std::vector<std::pair<std::size_t, double> > const& inputs = contents[i].second;
+		if(!inputs.empty())
+			maxIndex = std::max(maxIndex, inputs.back().first);
+	}
+	maxIndex = std::max<std::size_t>(maxIndex,dimensions);
```
Ulfgard: there should be an exception if maxIndex > dimensions if dimensions is not 0

include/shark/Data/SparseData.h
Ulfgard: agreed!

include/shark/Data/SparseData.h
TGlas: >>
>>   	size_t dim = inputDimension(dataset);
>>   	if(numberOfClasses(dataset)!=2) oneMinusOne = false;
>>
>>   	std::vector<detail::LabelSortPair> L;
>> -	if(sortLabels) {
>> +	if (sortLabels) {
>>   		for(std::size_t i = 0; i < elements; i++)
>>   			L.push_back(detail::LabelSortPair(dataset.element(i).label, i));
>
> When we are at it, LabelSortPair can be replaced by a shark::KeyValuePair<unsigned int,double> found in Core/utility/KeyValuePair.h and instead of the constructor one can use makeKeyValuePair(dataset.element(i).label, i)

I will replace it with a proper C++11 solution that also
gets along without element access. I see, there is more to
fix than the missing regression support.

src/Data/SparseData.cpp
TGlas: >>   inline std::vector<LibSVMPoint>
>> -importSparseDataReader(
>> -	std::istream& stream
>> -) {	
>> -	std::vector<LibSVMPoint>  fileContents;
>> -	while(stream){
>> +importSparseDataReader(std::istream& stream) {
>
> for VC++ it might help to put this into a separate compilation unit and add a header entry for it. Spirit takes a lot of ressources during compilation

Not sure what you mean. This is in a cpp file, not a header,
so we should be okay? The header knows nothing about spirit.

include/shark/Data/SparseData.h
```diff
@@ -173,39 +164,45 @@ SHARK_EXPORT_SYMBOL void importSparseData(
...
 
 	size_t dim = inputDimension(dataset);
 	if(numberOfClasses(dataset)!=2) oneMinusOne = false;
 
-	std::vector<detail::LabelSortPair> L;
-	if (sortLabels) {
-		for(std::size_t i = 0; i < elements; i++) 
-			L.push_back(detail::LabelSortPair(dataset.element(i).label, i));
-		std::sort (L.begin(), L.end(), detail::cmpLabelSortPair);
+	std::vector< std::pair<std::size_t, std::size_t> > order;
```
Ulfgard: using KeyLabelPair would have been much(!) easier, there you get the op< for free

include/shark/Data/SparseData.h
```diff
@@ -173,39 +164,45 @@ SHARK_EXPORT_SYMBOL void importSparseData(
...
+				}
+			);
 	}
 
-	for(std::size_t ii = 0; ii < elements; ii++) {
-		// apply mapping to sorted indices
-		std::size_t i = 0;
-		if(sortLabels) i = L[ii].second;
-		else i = ii;
+	for (std::size_t i=0; i<order.size(); i++)
```
Ulfgard: for(auto const& element: order)

Ulfgard:
looks good. go!

PR #72, opened 2016-04-19
added a fast algorithm for the HypSSP in 2D

Ulfgard:
Do not try to understand the paper.

This algorithm solves the problem of selecting the set of k points out of a set of n points which maximize the hypervolume over k points. This is only easily solvable in 2D, otherwise you face exponential run-time. At least by current state-of-the-art which has a O(n^k) factor in it because of the need to enumerate all possible sets.

### DIFF
--- a/include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
+++ b/include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
@@ -39,6 +39,12 @@ namespace shark {
 /// This algorithm solves the problem of selecting a subset of points with largest hypervolume in 2D.
 /// The algorithm has complexity n (k+log(n)) 
 ///
+/// While this algorithm accepts fronts with dominated points in it, the caller has to ensure
+/// that after domination checks there are at least as many points left as there are to select. The
+/// Algorithm will throw an exception otherwise.
+///
+/// This can easily be ensured by removing the nondominated points prior to calling this function.
+///
 /// The algorithm is described in:
 /// Bringmann, Karl, Tobias Friedrich, and Patrick Klitzke. 
 /// "Two-dimensional subset selection for hypervolume and epsilon-indicator."
@@ -238,6 +244,11 @@ private:
 	}
 public:
 	/// \brief Executes the algorithm.
+	/// While this algorithm in general accepts fronts with dominated points in it, the caller has to ensure
+	/// that after domination checks there are at least as many points left as there are to select. The
+	/// Algorithm will throw an exception otherwise.
+	///
+	/// This can easily be ensured by removing the nondominated points prior to calling this function.
 	/// \param [in] points The set \f$S\f$ of points to select
 	/// \param [out] selected set of the same size as the set of points indicating whether the point is selected (1) or not (0)
 	/// \param [in] k number of points to select. Must be lrger than 0
@@ -268,13 +279,18 @@ public:
 	/// \brief Executes the algorithm.
 	///
 	/// This version does not use a reference point. instead the extreme points are always kept which  implicitely defines a reference point
+	/// that after domination checks there are at least as many points left as there are to select. The
+	/// Algorithm will throw an exception otherwise.
+	///
+	/// This can easily be ensured by removing the nondominated points prior to calling this function.
+	///
 	/// \param [in] points The set \f$S\f$ of points to select
 	/// \param [out] selected set of the same size as the set of points indicating whether the point is selected (1) or not (0)
-	/// \param [in] k number of points to select, must be larger than 2
+	/// \param [in] k number of points to select, must be larger or equal 2
 	template<typename Set, typename SelectedSet>
 	void operator()( Set const& points, SelectedSet& selected, std::size_t k){
 		SIZE_CHECK(points.size() == selected.size());
-		SHARK_CHECK( k > 2, "[HypervolumeSubsetSelection2D] k must be larger than 2");
+		SHARK_CHECK( k >= 2, "[HypervolumeSubsetSelection2D] k must be larger or equal 2");
 		SHARK_CHECK( k <= points.size(), "[HypervolumeSubsetSelection2D] the number of points mjust be larger than k");
 		SIZE_CHECK(points.size() == selected.size());
 		SIZE_CHECK( points.begin()->size() == 2 );
@@ -299,6 +315,7 @@ public:
 		selected[front.back().index] = true;
 		front.pop_back();
 		front.erase(front.begin(),front.begin()+1);
+		if(k == 2) return;
 		
 		//find the optimal set in the front. afterwards selected points have selected=true
 		hypSSP(front,k-2);
--- a/include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
+++ b/include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
@@ -1,0 +1,298 @@
+/*!
+ *
+ * \author      O.Krause, T. Glasmachers
+ * \date        2014-2016
+ *
+ *
+ * \par Copyright 1995-2016 Shark Development Team
+ * 
+ * <BR><HR>
+ * This file is part of Shark.
+ * <http://image.diku.dk/shark/>
+ * 
+ * Shark is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published 
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * 
+ * Shark is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Shark.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+#ifndef SHARK_ALGORITHMS_DIRECTSEARCH_HYPERVOLUMESUBSETSELECTION_2D_H
+#define SHARK_ALGORITHMS_DIRECTSEARCH_HYPERVOLUMESUBSETSELECTION_2D_H
+
+#include <shark/LinAlg/Base.h>
+
+#include <algorithm>
+#include <vector>
+#include <deque>
+
+namespace shark {
+/// \brief Implementation of the exact hypervolume subset selection algorithm in 2 dimensions.
+///
+/// This algorithm solves the problem of selecting a subset of points with largest hypervolume in 2D.
+/// The algorithm has complexity n (k+log(n)) 
+///
+/// The algorithm is described in:
+/// Bringmann, Karl, Tobias Friedrich, and Patrick Klitzke. 
+/// "Two-dimensional subset selection for hypervolume and epsilon-indicator."
+/// Proceedings of the 2014 conference on Genetic and evolutionary computation. 
+/// ACM, 2014.	
+/// (although it is not very helpful)
+struct HypervolumeSubsetSelection2D {
+private:
+	
+	struct Point{
+		Point(){}
+
+		Point(double f1, double f2, std::size_t index)
+		: f1(f1)
+		, f2(f2)
+		, index(index)
+		, selected(false)
+		{}
+			
+		bool operator<(Point const& rhs) const{//for lexicographic sorting
+			if (f1 < rhs.f1) return true;
+			if (f1 > rhs.f1) return false;
+			return (f2 < rhs.f1);
+		}
+		
+		double f1;
+		double f2;
+		std::size_t index;
+		bool selected;
+	};
+	
+	///\brief Linear function a*x+b where a is stored in first and b is stored in section.
+	///
+	/// The linear function also stores an index to uniquely identify it
+	struct LinearFunction{
+	
+		double a;
+		double b;
+		std::size_t index;
+		
+		LinearFunction(double a, double b, std::size_t index = 0):a(a), b(b), index(index){}
+		LinearFunction(){}
+		
+		double eval(double x)const{
+			return a*x + b;
+		}
+	};
+
+	/// \brief Returns the intersection of two linear functions
+	double Intersection(LinearFunction f1, LinearFunction f2)const{
+		return (f2.b - f1.b) / (f1.a - f2.a);
+	}
+	
+	
+	/// \brief  Calculates for each given x the maximum among the functions f, i.e. the upper envelope of f.
+	/// 
+	/// Algoruthm 2 in the paper. Complexity O(n)
+	/// given a set of functions f_1...f_n, ordered by slope such that f_1.a < f_2.a<... and points with x-coordinate x_1<...<x_n
+	/// computes h_i = max_{1 <= j < i} f_j(x_i) for i=1,...,n as well as the index of the function leading to the value h_i
+	std::pair<std::vector<double>,std::vector<std::size_t> > upperEnvelope(
+		std::vector<LinearFunction>const& functions,
+		std::vector<Point> const& points
+	)const{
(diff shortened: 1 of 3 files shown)


include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
+/*!
+ *
+ * \author      O.Krause, T. Glasmachers
```
TGlas: I didn't do anything :)

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+			return (f2 < rhs.f1);
+		}
+		
+		double f1;
+		double f2;
+		std::size_t index;
+		bool selected;
+	};
+	
+	///\brief Linear function a*x+b where a is stored in first and b is stored in section.
```
TGlas: The documentation for this type is not very helpful. What's the purpose of the type? Well, compute a linear function. Please provide a bit more context.

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+		LinearFunction(){}
+		
+		double eval(double x)const{
+			return a*x + b;
+		}
+	};
+
+	/// \brief Returns the intersection of two linear functions
+	double Intersection(LinearFunction f1, LinearFunction f2)const{
+		return (f2.b - f1.b) / (f1.a - f2.a);
```
TGlas: We should probably assert or generally throw an exception if the denominator is zero, or better if the result is not finite.

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+	}
+	
+	
+	/// \brief  Calculates for each given x the maximum among the functions f, i.e. the upper envelope of f.
+	/// 
+	/// Algoruthm 2 in the paper. Complexity O(n)
```
TGlas: typo, should be Algorithm

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+	/// given a set of functions f_1...f_n, ordered by slope such that f_1.a < f_2.a<... and points with x-coordinate x_1<...<x_n
+	/// computes h_i = max_{1 <= j < i} f_j(x_i) for i=1,...,n as well as the index of the function leading to the value h_i
+	std::pair<std::vector<double>,std::vector<std::size_t> > upperEnvelope(
+		std::vector<LinearFunction>const& functions,
+		std::vector<Point> const& points
+	)const{
+		std::size_t n = points.size();
```
TGlas: Let's add SHARK_ASSERT(functions.size() == n);

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+			// at the end of s are the functions with largest slope.
+			// therefore if we have the last two elements as s_-1 and s_-2 and the new
+			// function f, knowing that the intersection of s_-1 and f is smaller than the intersection
+			// of s_-1 and s_-2 means that s_-1 is dominated  by s_-2 and f and thus can be removed.
+			while (s.size() > 1 ) {
+				
+				double d1 = Intersection(functions[i], s.end()[-1]);
+				double d2 = Intersection(s.end()[-2], s.end()[-1]);
+
+				if (d1 <= d2 || std::abs(d1-d2) < 1.e-10) {//check for numeric stability
```
TGlas: This can be simplified to the condition d1 <= d2 + 1e-10, but we can also leave this optimization to the compiler because the current code make the intent more clear.

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+	double Intersection(LinearFunction f1, LinearFunction f2)const{
+		return (f2.b - f1.b) / (f1.a - f2.a);
+	}
+	
+	
+	/// \brief  Calculates for each given x the maximum among the functions f, i.e. the upper envelope of f.
+	/// 
+	/// Algoruthm 2 in the paper. Complexity O(n)
+	/// given a set of functions f_1...f_n, ordered by slope such that f_1.a < f_2.a<... and points with x-coordinate x_1<...<x_n
+	/// computes h_i = max_{1 <= j < i} f_j(x_i) for i=1,...,n as well as the index of the function leading to the value h_i
```
TGlas: This description suggets that h_1 is undefined since the index set 1 <= j < i is empty for i=1. However, the implementation seems to use the index set 1 <= j <= i.

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+		return std::make_pair(std::move(h),std::move(chosen));
+	}
+	
+	
+	/// Fast calculation O(n*k) for the hypervolume selection problem
+	/// returns the selected points
+	void hypSSP(std::vector<Point>& front,std::size_t k)const{
+		std::size_t n = front.size();
+		std::vector<LinearFunction> functions(n);
+
```
TGlas: Here is a good place for SHARK_ASSERT(k > 0 && k < n);

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+	
+	/// Fast calculation O(n*k) for the hypervolume selection problem
+	/// returns the selected points
+	void hypSSP(std::vector<Point>& front,std::size_t k)const{
+		std::size_t n = front.size();
+		std::vector<LinearFunction> functions(n);
+		
+		std::vector<std::vector<std::size_t> > chosen;
+		std::vector<double>  h(n,0.0);
+		for(std::size_t j=0; j != k-1; ++j) {//compute until k-1 elements are chosen
```
TGlas: Replacing j != k-1 with j+1 != k is not subject to underflow (but it does not solve the problem if k==0)

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+			// the function with the largest function value is currently at the front
+			h[i] = s[0].eval(points[i].f1);
+			chosen[i] = s[0].index;
+		}
+		return std::make_pair(std::move(h),std::move(chosen));
+	}
+	
+	
+	/// Fast calculation O(n*k) for the hypervolume selection problem
+	/// returns the selected points
```
TGlas: Well, no. The function does not return the selected points. It merely marks the points as selected.

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+	void hypSSP(std::vector<Point>& front,std::size_t k)const{
```
TGlas: After reading the code I have no idea how and why the function works. Okay, I should try to read the paper first :)

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+	template<typename Set>
+	std::vector<Point> createFront(Set const& points, double refX, double refY)const{
+		//copy points using the new reference frame with refPoint at (0,0). also store original index for later
+		std::vector<Point> front;
+		for(std::size_t i = 0; i != points.size(); ++i){
+			front.emplace_back(points[i](0) - refX, points[i](1) - refY,i);
+		}
+		std::sort(front.begin(),front.end());//sort lexicographically
+		//erase dominated points
+		auto newEnd = std::unique(front.begin(),front.end(),[](Point const& x, Point const& y){
```
TGlas: Without the comments this would be entirely crypric, but with the comments in place it is all clear, and actually really neat!

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+	/// \param [in] k number of points to select
+	/// \param [in] refPoint The reference point \f$\vec{r} \in \mathbb{R}^2\f$ for the hypervolume calculation, needs to fulfill: \f$ \forall s \in S: s \preceq \vec{r}\f$. .
+	template<typename Set, typename SelectedSet, typename VectorType >
+	void operator()( Set const& points, SelectedSet& selected, std::size_t k, VectorType const& refPoint){
+		SIZE_CHECK(points.size() == selected.size());
+		if(points.empty())
+			return;
+		SIZE_CHECK( points.begin()->size() == 2 );
+		SIZE_CHECK( refPoint.size() == 2 );
+
```
TGlas: SHARK_CHECK( k > 0, "[HypervolumeSubsetSelection2D] k must be non-zero");

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+	/// \param [out] selected set of the same size as the set of points indicating whether the set is selected (1) or not (0)
+	/// \param [in] k number of points to select, must be larger than 2
+	template<typename Set, typename SelectedSet, typename VectorType >
+	void operator()( Set const& points, SelectedSet& selected, std::size_t k){
+		SIZE_CHECK(points.size() == selected.size());
+		if(points.empty())
+			return;
+		SIZE_CHECK( points.begin()->size() == 2 );
+		SHARK_CHECK( k > 2, "[HypervolumeSubsetSelection2D] k must be larger than 2");
+
```
TGlas: We need the condition points.size() >= 2, right? We could also allow points.size() == 1 and k == 1, which means that the point is kept. Maybe more important, we should check k <= points.size().

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
```diff
@@ -0,0 +1,298 @@
...
+	/// \param [in] points The set \f$S\f$ of points to select
+	/// \param [out] selected set of the same size as the set of points indicating whether the set is selected (1) or not (0)
+	/// \param [in] k number of points to select, must be larger than 2
+	template<typename Set, typename SelectedSet, typename VectorType >
+	void operator()( Set const& points, SelectedSet& selected, std::size_t k){
+		SIZE_CHECK(points.size() == selected.size());
+		if(points.empty())
+			return;
+		SIZE_CHECK( points.begin()->size() == 2 );
+		SHARK_CHECK( k > 2, "[HypervolumeSubsetSelection2D] k must be larger than 2");
```
TGlas: What's wrong with k==2 ?

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
Ulfgard: obviously k must be > 0. i add this

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
Ulfgard: what if n>k points are given but after removing the dominating points we have front.size()<k? exception?

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
Ulfgard: this would be trivial, right? pick the edge points and be done. The algorithm can not really do its job as the set does not containa  single point that is actually picked by hypervolume.

TGlas: >> +		}
>> +	}
>> +	
>> +	/// \brief Executes the algorithm.
>> +	///
>> +	/// This version does not use a reference point. instead the extreme points are kept which defines a reference point
>> +	/// \param [in] points The set \f$S\f$ of points to select
>> +	/// \param [out] selected set of the same size as the set of points indicating whether the set is selected (1) or not (0)
>> +	/// \param [in] k number of points to select, must be larger than 2
>> +	template<typename Set, typename SelectedSet, typename VectorType >
>> +	void operator()( Set const& points, SelectedSet& selected, std::size_t k){
>> +		SIZE_CHECK(points.size() == selected.size());
>> +		if(points.empty())
>> +			return;
>> +		SIZE_CHECK( points.begin()->size() == 2 );
>> +		SHARK_CHECK( k > 2, "[HypervolumeSubsetSelection2D] k must be larger than 2");
>
> this would be trivial, right? pick the edge points and be done. The algorithm can not really do its job as the set does not containa  single point that is actually picked by hypervolume.

Yes, it is a trivial corner case. But why should it assert? It should 
simply return the edge points (and succeed).

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeSubsetSelection2D.h
TGlas: >> +	}
>> +	
>> +	/// \brief Executes the algorithm.
>> +	///
>> +	/// This version does not use a reference point. instead the extreme points are kept which defines a reference point
>> +	/// \param [in] points The set \f$S\f$ of points to select
>> +	/// \param [out] selected set of the same size as the set of points indicating whether the set is selected (1) or not (0)
>> +	/// \param [in] k number of points to select, must be larger than 2
>> +	template<typename Set, typename SelectedSet, typename VectorType >
>> +	void operator()( Set const& points, SelectedSet& selected, std::size_t k){
>> +		SIZE_CHECK(points.size() == selected.size());
>> +		if(points.empty())
>> +			return;
>> +		SIZE_CHECK( points.begin()->size() == 2 );
>> +		SHARK_CHECK( k > 2, "[HypervolumeSubsetSelection2D] k must be larger than 2");
>> +		
>
> what if n>k points are given but after removing the dominating points we have front.size()<k? exception?

Oh, good point. The use case is selection, therefore (at first glance) 
an exception does not make much sense here. I see two possible solutions:

1. The selection operator performs non-domianted sorting and then calls 
this algorithm on the non-dominance rank that must be split into 
selected and discarded individuals. This removes the need to check 
dominated points, and then throwing an exception is a good solution.

2. The selection operator calls this function directly. If we find fewer 
than k non-dominated points then we have to emulate the above behavior 
by accepting all non-dominated points and continuing with the rest.

In any case the documentation of the function needs an update to reflect 
the new behavior.

Ulfgard:
I incorporated the suggestions and also ensured that the unit tests cover everything.

TGlas:
Looks all good, ready for merge.

PR #73, opened 2016-04-21
Added initial interface for hypervolume contribution algorithms

Ulfgard:
This pull request is mostly there to start an initial code review and discussion on the design on an interface for the contribution algorithms. 

I went for the route which returns the indices of the k-least or k largest contributors with and without reference point.

I have added a 3D-Version of the algorithm.

### DIFF
--- a/include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContribution2D.h
+++ b/include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContribution2D.h
@@ -1,0 +1,173 @@
+/*!
+ *
+ * \author      O.Krause, T. Glasmachers
+ * \date        2014-2016
+ *
+ *
+ * \par Copyright 1995-2016 Shark Development Team
+ * 
+ * <BR><HR>
+ * This file is part of Shark.
+ * <http://image.diku.dk/shark/>
+ * 
+ * Shark is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published 
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * 
+ * Shark is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Shark.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+#ifndef SHARK_ALGORITHMS_DIRECTSEARCH_HYPERVOLUME_CONTRIBUTION_2D_H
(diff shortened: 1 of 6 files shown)


include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContribution2D.h
```diff
@@ -0,0 +1,173 @@
...
+		std::vector<std::pair<double,std::size_t> > bestK(k+1);
+		auto heapStart = bestK.begin();
+		auto heapEnd = bestK.begin();
+		
+		auto pointComp = [&](std::pair<double,std::size_t> const& lhs, std::pair<double,std::size_t> const& rhs){return comp(lhs.first,rhs.first);};
+		
+		//compute the hypervalue contribution for each point except the endpoints
+		*heapEnd = std::make_pair((front[2].f1 - front[1].f1) * (front[0].f2 - front[1].f2),front[1].index);
+		++heapEnd;
+		for(std::size_t i = 2; i < front.size()-1;++i){
```
TGlas: Hmmm... if you want all but the extremes, shouldn't the loop start at i = 1 ?

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContribution2D.h
```diff
@@ -0,0 +1,173 @@
...
+		
+		return bestContributors(front,k,[](double con1, double con2){return con1 < con2;});
+	}
+	
+	/// \brief Returns the index of the points with largest contribution.
+	///
+	/// \param [in] points The set \f$S\f$ of points from which to select the least contributor.
+	/// \param [in] referencePoint The reference point \f$\vec{r} \in \mathbb{R}^2\f$ for the hypervolume calculation, needs to fulfill: \f$ \forall s \in S: s \preceq \vec{r}\f$.
+	template<typename Set, typename VectorType>
+	std::vector<std::size_t> largest( Set const& points, std::size_t k, VectorType const& referencePoint)const{
```
TGlas: "largest" is the opposite of "smallest", "most" is the opposite of "least". Another possible pair would be worst/best, which is probably closest to what we want here. So I propose to rename least -> worst and largest -> best.

TGlas:
Where exactly do you see a need for discussing the interface design? It is probably not just the names, right?

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContribution2D.h
Ulfgard: no, the point i=1 is already inserted in line 82. hm maybe i can unify this...yeah probably

Ulfgard:
for example: do we also want the contribution itself?

TGlas:
> for example: do we also want the contribution itself?

This is for selection in an EA. Right now I don't foresee any other 
potential applications (at least, within Shark). For this purpose the 
contribution does not matter much.

However, in other contexts the code could be used for performance 
monitoring and analysis (e.g., in BBOB). Then the size of the 
contribution may matter. Since it is anyway available, why not make it 
accessible?

Ulfgard:
is this now okay?

TGlas:
Yes, thanks, great!

PR #77, opened 2016-05-28
Multi objective function

Ulfgard:
Generates a multi-objective function from a set of single objective function by C++11 type-magic.

This change also includes a change in AbstractObjectiveFunction to have the proper type for multi-objective derivatives. Thus, this Benchmark also includes the derivatives.

(I had to close the old pull-request because i was not able to properly rebase for some reason)


include/shark/ObjectiveFunctions/AbstractObjectiveFunction.h
```diff
@@ -83,9 +87,15 @@ class AbstractObjectiveFunction :  public INameable{
 	typedef PointType SearchPointType;
 	typedef ResultT ResultType;
 
-	typedef SearchPointType FirstOrderDerivative;
+	//if the result type is not an arithmetic type, we assume it is a vector-type->multi objective optimization
+	typedef typename boost::mpl::if_<
+		std::is_arithmetic<ResultT>,
+		SearchPointType,
```
TGlas: If would look more natural to me to replace SearchPointType with RealVector. Well, this is probably not the place for a CompressedRealVector, and a derivative does not make any sense on a discrete search space.

Ulfgard: Hi,

this does not really hurt being there and it allows us in the future to look at different differentiable spaces, e.g. complex.

include/shark/ObjectiveFunctions/Benchmarks/MultiObjectiveBenchmark.h
```diff
@@ -0,0 +1,189 @@
...
+#include <shark/ObjectiveFunctions/BoxConstraintHandler.h>
+#include <shark/Rng/GlobalRng.h>
+
+#include <shark/LinAlg/rotations.h>
+#include <tuple>
+
+namespace shark {
+
+namespace detail{
+	//taken from the web. implements an std::integer_sequence type representing a sequence 0,...,N-1, std::integer_sequence is not here until C++14
```
TGlas: Does "taken from the web" mean that there is a license somewhere, of can we safely assume "public domain"?

include/shark/ObjectiveFunctions/Benchmarks/MultiObjectiveBenchmark.h
```diff
@@ -0,0 +1,189 @@
...
+	///
+	/// The name has the form Objective1/Objective2/Objective3/.../ObjectiveN
+	/// where ObjectiveK is the name of the k-th objective.
+	std::string name()const{
+		return generateName(typename detail::generate_integer_sequence<sizeof...(Objectives)>::type());
+		
+	}
+	
+	bool hasScalableDimensionality()const{
+		return true;
```
TGlas: This does not make much sense to me. Shouldn't this function ask the underlying single objective functions for this property? Alternatively it can assert this in the constructor.

include/shark/ObjectiveFunctions/Benchmarks/MultiObjectiveBenchmark.h
```diff
@@ -0,0 +1,189 @@
...
+	}
+	
+	void setNumberOfVariables( std::size_t numberOfVariables ){
+		for(auto& f: m_rotations){
+			f.setNumberOfVariables(numberOfVariables);
+		}
+	}
+	
+	std::size_t numberOfObjectives()const{
+		return m_rotations.size();
```
TGlas: Hm funny, why use a runtime value? Can't we return sizeof...(Objectives)? Actually, I don't have any experience with this construction.

include/shark/ObjectiveFunctions/Benchmarks/MultiObjectiveBenchmark.h
```diff
@@ -0,0 +1,189 @@
...
+	
+	///\ Initializes the functions as well as picks random rotations and translations
+	void init() {
+		m_translations.clear();
+		
+		for(auto& f: m_rotations)
+		{
+			RealVector translation(numberOfVariables());
+			for(double& v: translation){
+				v=Rng::gauss(0,1)/numberOfVariables();
```
TGlas: Shouldn't this divide by sqrt(numberOfVariables()) to obtain an expected squared norm of one?

include/shark/ObjectiveFunctions/Benchmarks/MultiObjectiveBenchmark.h
```diff
@@ -0,0 +1,189 @@
...
+			f.init();
+		}
+	}
+	
+	SearchPointType proposeStartingPoint() const {
+		RealVector x(numberOfVariables());
+
+		std::size_t index = Rng::discrete(0,m_rotations.size()-1);
+		for (std::size_t i = 0; i < x.size(); i++) {
+			x(i) = m_translations[index](i)+Rng::gauss(0,1)/numberOfVariables();
```
TGlas: Same as above, missing sqrt?

TGlas:
Overall the code makes perfect sense. All of the above comments are minor. Just fix and commit.

include/shark/ObjectiveFunctions/Benchmarks/MultiObjectiveBenchmark.h
Ulfgard: stackoverflow. so this is safe, probably also the standard implementation since c++14

include/shark/ObjectiveFunctions/Benchmarks/MultiObjectiveBenchmark.h
Ulfgard: this should give an exception in the constructor anyways. but i added a meaningful check to make this more clear. As this is a benchmark function it does not really make sense to have fixed dimensionality. Otherwise i would need to find a way to construct the objects in a way that all end up with the same dimensionality.

include/shark/ObjectiveFunctions/Benchmarks/MultiObjectiveBenchmark.h
Ulfgard: fixed (and the next)

PR #79, opened 2016-06-03
Hypervolume algorithms and indicators

Ulfgard:
I reworked this once again.

New things: 

1. cleaned up all log-volume and multiplicative-epsilon stuff. This can easier be done using a log-transformation of the objective function and thus is far less error prone. => removed

2. Removed the LeastContributorApproximator and added it to the HypervolumeContribution family of algorithms in Operators/Hypervolume. Code-Review of that part is highly(!) appreciated. I am not sure whether everything is fully working as intended. I removed the upper bound on the samples as the algorithm is guarantueed to converge. For now i also removed the exact calculation step because this would make testing of it much harder(and it is much cleaner now). Can be added again if required.  

3. HypervolumeIndicator now uses Operators/Hypervolume/HypervolumeContribution.h as backend. It allows to set a reference point explicitly, if it is done than contribution is computed on *all* points. Otherwise, the version without reference point is called which estimates the reference by the maximum value found in the set and then removing the extremum points from the points to compute the contributions on.

4. All Algorithms now have an indicator() member function which returns the current indicator for configuration purposes. Via this the reference point can be set and the approximation algorithm configured

5. Indicators were moved to Operators/Indicators


Test/Algorithms/DirectSearch/Operators/HypervolumeContribution.cpp
```diff
@@ -184,4 +190,74 @@ BOOST_AUTO_TEST_CASE( Algorithms_HypervolumeContribution3D ) {
...
+		std::vector<double> approxContributions;
+		for(std::size_t i = 0; i != numTrials; ++i){
+			auto result = algorithm.smallest(set,1,reference)[0].value;
+			approxContributions.push_back(contributions[result]);
+		}
+		std::sort(approxContributions.begin(),approxContributions.end());
+		
+		//check that we do not have too many errors, i.e. contributions with errors larger than 1+epsilon
+		//we make on average 100*errorProbability=10 errors. we give 50% more slack
+		BOOST_CHECK_LT(approxContributions[(1-1.5*algorithm.delta())*numTrials], (1+algorithm.epsilon())*contributionsTrue[0].key);
```
TGlas: Hmmm... that's not a good test. In the worst case the number of failures is a binomial distribution with p=0.1. Hence the variance is np(1-p) = 9, stddev=3, and the slack of 5 is not even two standard deviations.
1. besser express the slack as a multiple of the standard deviation,
2. better use three standard deviations to be on the safe side.
This shifts the threshold from 15 to 19, which is still a meaningful test.

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContribution.h
```diff
@@ -0,0 +1,176 @@
...
+	void serialize( Archive & archive, const unsigned int version ) {
+		archive & BOOST_SERIALIZATION_NVP(m_useApproximation);
+		archive & BOOST_SERIALIZATION_NVP(m_approximationAlgorithm);
+	}
+	
+	/// \brief Returns the index of the points with smallest contribution as well as their contribution.
+	///
+	/// \param [in] points The set \f$S\f$ of points from which to select the smallest contributor.
+	/// \param [in] k The number of points to select.
+	/// \param [in] referencePointThe reference Point\f$\vec{r} \in \mathbb{R}^2\f$ for the hypervolume calculation, needs to fulfill: \f$ \forall s \in S: s \preceq \vec{r}\f$.
```
TGlas: Space missing between "referencePoint" and "The" - please fix, since this is of syntactic importanct to doxygen :)

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContributionApproximator.h
```diff
@@ -0,0 +1,397 @@
...
+	/// \param [in] s pareto front of points
+	/// \param [in] reference The reference point to consider for calculating individual points' contributions.
+	template<class Set,class VectorType>
+	std::vector<KeyValuePair<double,std::size_t> > smallest(Set const& points, std::size_t k, VectorType const& reference)const{
+		SHARK_CHECK(k == 1, "not implemented for k != 1");
+				
+		std::vector< Point<VectorType> > front;
+		for(auto it = points.begin(); it != points.end(); ++it ) {
+			front.emplace_back( *it, reference );
+		}
```
TGlas: Can't this loop be written without iterators? Not sure because Set is a template argument, but it seems to me that the loop is exactly equivalent to:

    for (auto const& point : points) front.emplace_back(point, reference);

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContributionApproximator.h
```diff
@@ -0,0 +1,397 @@
...
+				return *minimalElement;
+			}
+			
+			delta *= m_multiplierDelta;
+		}
+	}
+	
+	/// \brief Samples in the bounding box of the supplied point until a pre-defined threshold is reached.
+	///
+	/// \param [in] s Set of points.
```
TGlas: The parameter "s" was apparently removed, but its documentation was not.

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContributionApproximator.h
```diff
@@ -0,0 +1,397 @@
...
+	/// \param [in] point Iterator to the point that should be sampled.
+	/// \param [in] r The current round.
+	/// \param [in] delta The delta that should be reached.
+	template<class VectorType>
+	void sample( Point<VectorType>& point, unsigned int r, double delta, std::size_t n )const{
```
TGlas: Parameter n is not explained in the documentation.

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContributionApproximator.h
```diff
@@ -0,0 +1,397 @@
...
+		//this is formula (3) in the paper when used in an equality < delta and solving for noSamples
+		//we add +1 to ensure that the inequality holds.
+		double thresholdD= 1.0+0.5 * ( (1. + m_gamma) * logR + logFactor ) * sqr( point.boundingBoxVolume / delta );
+		std::size_t threshold = static_cast<std::size_t>(thresholdD);
+		//sample points inside the box of the current point
+		for( ; point.noSamples < threshold; point.noSamples++ ) {
+			//sample a point inside the box
+			point.sample.resize(point.point.size());
+			for( unsigned int i = 0; i < point.sample.size(); i++ ) {
+				point.sample[ i ] = point.point[ i ] + Rng::uni( 0., 1. ) * ( point.boundingBox[ i ] - point.point[ i ] );
```
TGlas: I think that this line should read

    point.sample[ i ] = Rng::uni(point.point[ i ], point.boundingBox[ i ]);

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContributionMD.h
```diff
@@ -0,0 +1,179 @@
...
+		double baseVol = hv(points,ref);
+		std::vector<KeyValuePair<double,std::size_t> > result;
+		SHARK_PARALLEL_FOR( int i = 0; i < static_cast< int >( points.size() ); i++ ) {
+			if(std::find(minIndex.begin(),minIndex.end(),i) != minIndex.end())
+				continue;
+			std::vector<RealVector> copy( points.begin(), points.end() );
+			copy.erase( copy.begin() + i );
+			
+			double volume = baseVol - hv(copy,ref);
+			SHARK_CRITICAL_REGION{
```
TGlas: Can we get around using a critical section by pre-allocating space for the vector? Maybe I am missing the purpose of this.

TGlas:
The changes all look good to me, a few details are in the comments above.

I tried my best to check the least-contribution-approximator code. I makes sence, but to properly check its correctness I would have to work through the paper again. The general proceeding (the sampling) is straightforward, and I trust the implementation. However, I cannot recall how the "influencing points" are supposed to be handled.

include/shark/Algorithms/DirectSearch/Operators/Hypervolume/HypervolumeContributionMD.h
Ulfgard: the problem at this point is that we skip the extrema. if we pre-allocate, we need to find out how many points we already skipped in iteration i or later remove those points again. I figured that the critical region+ push_back is the simplest solution. It does not really hurt, though as the cost of computing the contribution is a few orders of magnitudes larger.

Ulfgard:
the influencing points are introduced in one of the last chapters of the paper where they create the practical algorithm. The idea is to remove all points which cover some part of the bounding box but that volume is already completely covered by some other point in the set. Our current algorithm uses a cheap approximation for it. we simply shrink the bounding box when we find that it sliced through by some other BB in one direction. the influencing points are then points that partly intersect with the box.

the same trick can be applied to the exact computation. another trick is to project the influencing points onto the box. This allows us to check whether a point is dominated.

this is one of the main tricks used for the current fastest algorithm(in fact the currently fastest algorithm for exact hypervolume relies on a fast contribution algorithm). there is still work to be done.

PR #84, opened 2016-06-19
added the SAG stochastic gradient descent method

Ulfgard:
I added the SAG method for linear models, see

https://hal.inria.fr/hal-00860051/document

it has linear convergence rate on strongly convex functions, otherwise O(1/T). I could not find its convergence rate on non-smooth functions(if it converges at all).

The implementation works for arbitrary data and loss-functions, especially dense & sparse vectors, classification & regression and weighted & un-weighted.

### DIFF
--- a/Test/Algorithms/Trainers/LinearSAGTrainer.cpp
+++ b/Test/Algorithms/Trainers/LinearSAGTrainer.cpp
@@ -1,0 +1,191 @@
+//===========================================================================
+/*!
+ * 
+ *
+ * \brief       test case for the Linear SAG-Trainer
+ * 
+ * 
+ * 
+ *
+ * \author      O.Krause
+ * \date        2016
+ *
+ *
+ * \par Copyright 1995-2015 Shark Development Team
+ * 
+ * <BR><HR>
+ * This file is part of Shark.
+ * <http://image.diku.dk/shark/>
+ * 
+ * Shark is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published 
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * 
+ * Shark is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Shark.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+//===========================================================================
+#define BOOST_TEST_MODULE ALGORITHMS_TRAINERS_Linear_SAG_Trainer
+#include <boost/test/unit_test.hpp>
+#include <boost/test/floating_point_comparison.hpp>
+
+#include <shark/Algorithms/Trainers/LinearSAGTrainer.h>
+#include <shark/ObjectiveFunctions/Loss/CrossEntropy.h>
+#include <shark/ObjectiveFunctions/Loss/SquaredLoss.h>
+#include <shark/ObjectiveFunctions/ErrorFunction.h>
+#include <shark/ObjectiveFunctions/Regularizer.h>
+#include <shark/Data/DataDistribution.h>
+#include <shark/Statistics/Distributions/MultiVariateNormalDistribution.h>
+
+using namespace shark;
+
+BOOST_AUTO_TEST_SUITE (Algorithms_Trainers_Linear_SAG_Trainer)
+
+
+template<class Dataset>
+void testClassification(Dataset const& dataset, double lambda, unsigned int epochs, bool trainOffset){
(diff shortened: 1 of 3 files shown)


Test/Algorithms/Trainers/LinearSAGTrainer.cpp
```diff
@@ -0,0 +1,191 @@
...
+ * \brief       test case for the Linear SAG-Trainer
+ * 
+ * 
+ * 
+ *
+ * \author      O.Krause
+ * \date        2016
+ *
+ *
+ * \par Copyright 1995-2015 Shark Development Team
```
TGlas: 2015 -> 2016
We should really have a yearly cron job somewhere to automate this.

include/shark/Algorithms/Trainers/LinearSAGTrainer.h
```diff
@@ -0,0 +1,276 @@
+//===========================================================================
+/*!
+ *
+ *
+ * \brief       Generic Stochastic Average Gradient Descent training for linear models
+ *
+ *
+ *
+ *
+ * \author      T. Glasmachers
```
TGlas: Is it me? copy&paste...? :)

TGlas:
This is great, I really love the SAG algorithm!
There is only one detail, the loop in the trainer is terminated based on the number of epochs. Can't we replace that with an AbstractStoppingCriterion?

Ulfgard:
Hi,

the number of epochs is one of two possible stopping criteria in this case: number of epochs or change of point. as the change in function value or in gradient is an incredible noisy (and biased) statistic they do not make much sense. Change of point does not tell much about change of function value in ill-conditioned problems, so epochs it is.

Also this is the same stopping criterion as for the KernelSGDTrainer and I tried to match interfaces as close as possible.

should I also add a KernelSAGTrainer? thats straight forward to add.

Do you know any reference telling about convergence in the non-smooth case? I would like to add a comment about this.

TGlas:
> the number of epochs is one of two possible stopping criteria in this case: number of epochs or change of point. as the change in function value or in gradient is an incredible noisy (and biased) statistic they do not make much sense. Change of point does not tell much about change of function value in ill-conditioned problems, so epochs it is.
Okay, I see.

> Also this is the same stopping criterion as for the KernelSGDTrainer and I tried to match interfaces as close as possible.
Makes sense.

> should I also add a KernelSAGTrainer? thats straight forward to add.
I already thought about doing this myself :)
In my experience a "brute force" kernel trainer does not make too
much sence, but it is straightforward. If possible then please keep
the set of support vectors independent of the training data.
Then we can throw in the training data as a special case (standard
kernel setting), but we can also use a random projection / budget
approach.

> Do you know any reference telling about convergence in the non-smooth case? I would like to add a comment about this.
No. As far as I remember they simply don't say anything about this
case in the papers.

Look at this example problem (with optimum x^* = 0):
    min |x+1| + |x| + |x-1|
Let's assume that we have already converged to the area -1 <= x <= +1,
then the agerage gradient coincides with the gradient of |x|. So why
should it be better than plain SGD?

Practically speaking, with hinge loss it does not seem to converge
any faster than SGD, so for an SVM one should really use a squared
hinge or Huber-type loss.

Ulfgard:
> I already thought about doing this myself :) In my experience a "brute force" kernel trainer does not make too much sence, but it is straightforward. If possible then please keep the set of support vectors independent of the training data. Then we can throw in the training data as a special case (standard kernel setting), but we can also use a random projection / budget approach.

I thought about a version that is consistent with the kernel-SGD trainer. i.e. take the trainer and just replace the SGD part by SAG. So whatever we add, we should also add it to that trainer (or remove/replace/amend it). This means that we have the 2-norm regularization of the alpha-values instead of norm in hilbert-space, etc. If you want an independent basis, i can add that too. 

for the case of the L2-hinge loss we can add a "nice" caching strategy: only cache entries which have nonzero gradD components (they do not violate the margin -> gradient 0 and are also likely to be correct later on)

TGlas:
>> I already thought about doing this myself :) In my experience a "brute force" kernel trainer does not make too much sence, but it is straightforward. If possible then please keep the set of support vectors independent of the training data. Then we can throw in the training data as a special case (standard kernel setting), but we can also use a random projection / budget approach.
>
> I thought about a version that is consistent with the kernel-SGD trainer. i.e. take the trainer and just replace the SGD part by SAG. So whatever we add, we should also add it to that trainer (or remove/replace/amend it). This means that we have the 2-norm regularization of the alpha-values instead of norm in hilbert-space, etc. If you want an independent basis, i can add that too.

That would be great, I believe that this is a must for large-scale
learning.

> for the case of the L2-hinge loss we can add a "nice" caching strategy: only cache entries which have nonzero gradD components (they do not violate the margin -> gradient 0 and are also likely to be correct later on)

Yes. This is analogous to shrinking in the dual case, and you could
even do hard shrinking. Alternatively you can select those guys less
frequently. In my experience this can help, but it can also destroy
some of the stability of the method.
Exploiting the structure "only" at the level of the kernel cache
sounds like a good compromise.

Tobias

PR #85, opened 2016-06-22
Unified CSvm with MCSvm for easier interface

Ulfgard:
One problem of our interface was that we had a trainer for the binary case, 9 for the Mc case and this made life unneccesary hard (and confusing). So I decided to move all the McSvm trainers in an implementation folder and reference them via the CSvm class. I did the same for the Linear solver which allowed me to remove a few classes.

the new interface works like:

CSvmTrainer<RealVector> trainer;
trainer.setMcSvmType(McSVM::WW);//this is also the default choice.
trainer.train(model,data);//no more checking how many classes data have.

I updated documentation where available (we need way more). tests pass, examples give identical results.

### DIFF
--- a/include/shark/Algorithms/QP/BoxConstrainedProblems.h
+++ b/include/shark/Algorithms/QP/BoxConstrainedProblems.h
@@ -59,9 +59,8 @@ struct WS2MaximumGradientCriterion{
 				j = a;
 			}
 			if(secondLargestGradient > largestGradient){
-				using namespace std;
-				swap(secondLargestGradient,largestGradient);
-				swap(i,j);
+				std::swap(secondLargestGradient,largestGradient);
+				std::swap(i,j);
 			}
 		}
 		if(secondLargestGradient == 0)


TGlas:
Great, please go ahead and merge!

Ulfgard:
I have done some more changes and totally removed the old classes. please check whether this still looks alright.

include/shark/Algorithms/Trainers/CSvmTrainer.h
```diff
@@ -42,10 +38,10 @@ enum class McSvm{
 
 
 ///
-/// \brief Training of C-SVMs for binary classification.
+/// \brief Training of C-SVMs for binary classifellation.
```
TGlas: Awesome... I don't know this word, but all results on google are porn :)
This is apparently an artifact of the replacement "ic" -> "ell". It appears multiple times in the document.

include/shark/Algorithms/Trainers/CSvmTrainer.h
```diff
@@ -60,7 +56,7 @@ enum class McSvm{
 /// \f[
 ///     f(x) = \langle w, \phi(x) \rangle + b
 /// \f]
-/// with coefficients w and b given by the (primal)
+/// with coeffellients w and b given by the (primal)
```
TGlas: This time no porn...

include/shark/Algorithms/Trainers/CSvmTrainer.h
```diff
@@ -71,7 +67,7 @@ enum class McSvm{
 /// For details refer to the paper:<br/>
 /// <p>Support-Vector Networks. Corinna Cortes and Vladimir Vapnik,
 /// Machine Learning, vol. 20 (1995), pp. 273-297.</p>
-/// or simply to the Wikipedia article:<br/>
+/// or simply to the Wikipedia artellle:<br/>
```
TGlas: artellle -> article

include/shark/Algorithms/Trainers/CSvmTrainer.h
```diff
@@ -93,7 +89,7 @@ class CSvmTrainer : public AbstractSvmTrainer<
 	/// this and many of the below typedefs build on the class template type CacheType.
 	/// Simply changing that one template parameter CacheType thus allows to flexibly
 	/// switch between using float or double as type for caching the kernel values.
-	/// The default is float, offering sufficient accuracy in the vast majority
+	/// The default is float, offering suffellient accuracy in the vast majority
```
TGlas: guess what?

include/shark/Algorithms/Trainers/CSvmTrainer.h
```diff
@@ -102,7 +98,7 @@ class CSvmTrainer : public AbstractSvmTrainer<
 	typedef AbstractKernelFunction<InputType> KernelType;
 
 	//! Constructor
-	//! \param  kernel         kernel function to use for training and prediction
+	//! \param  kernel         kernel function to use for training and predelltion
```
TGlas: another one

include/shark/Algorithms/Trainers/CSvmTrainer.h
```diff
@@ -158,36 +155,82 @@ class CSvmTrainer : public AbstractSvmTrainer<
 				f.sparsify();
 			return;
 		}
-		//multiclass case: dispatch to the chosen Svm-type
+		//multelllass case: find correct dual formulation
```
TGlas: multiclass

TGlas:
After letting this rest for a month, I just spent more than half an hour trying to understand what you did. It turns out that the changes are just too many, and they are quite non-local. I kind-of get the big picture, but I dont't understand enough for a proper code review. Should we have a skype call, or do you feel confident to commit the code anyway?

Ulfgard:
skype call!

TGlas:
Instead of writing even more comments I have made a few really minor changes. The code looks good to me, and the integration into a single training is really nice. Please go ahead and merge, afterwards I'll have a closer look at issue #103.

Ulfgard:
i think the issue in your comment is that maybe std::swap was not declare dat that point... so an include of <utility> should have fixed it. still weird. I will commit tomorrow morning

TGlas:
> i think the issue in your comment is that maybe std::swap was not declare dat that point... so an include of <utility> should have fixed it. still weird. I will commit tomorrow morning

Indeed. Why then would it find the explicitly qualified std::swap?
I leared that doing it this way was good style, anyway, for integers
if does not matter at all.

PR #86, opened 2016-06-23
Added support for warm-starts into the CSvmTrainer.

TGlas:
I added warm start capabilities to the CSvmTrainer. This is needed urgently by a Masters student. Therefore I did not touch the other SVM trainers for now.

### DIFF
--- a/include/shark/Algorithms/QP/BoxConstrainedProblems.h
+++ b/include/shark/Algorithms/QP/BoxConstrainedProblems.h
@@ -273,6 +273,49 @@ public:
 	void reshrink(){}
 	void unshrink(){}
 
+	/// \brief Define the initial solution for the iterative solver.
+	///
+	/// This method can be used to warm-start the solver. It requires a
+	/// feasible solution (alpha) and the corresponding gradient of the
+	/// dual objective function.
+	void setInitialSolution(RealVector const& alpha, RealVector const& gradient)
+	{
+		std::size_t n = dimensions();
+		SIZE_CHECK(alpha.size() == n);
+		SIZE_CHECK(gradient.size() == n);
+		for (std::size_t i=0; i<n; i++)
+		{
+			SHARK_ASSERT(alpha(i) >= boxMin(i) && alpha(i) <= boxMax(i));
+			std::size_t j = permutation(i);
+			m_problem.alpha(i) = alpha(j);
+			m_gradient(i) = gradient(j);
+			m_alphaStatus[i] = ((alpha(i) == boxMin(i)) ? AlphaLowerBound : 0)
+			                 | ((alpha(i) == boxMax(i)) ? AlphaUpperBound : 0);
+		}
+	}
+
+	/// \brief Define the initial solution for the iterative solver.
+	///
+	/// This method can be used to warm-start the solver. It requires a
+	/// feasible solution (alpha), for which it computes the gradient of
+	/// the dual objective function. Note that this is a quadratic time
+	/// operation in the number of non-zero coefficients.
+	void setInitialSolution(RealVector const& alpha)
+	{
+		std::size_t n = dimensions();
+		SIZE_CHECK(alpha.size() == n);
+		RealVector gradient = m_problem.linear;
+		std::vector<QpFloatType> q(n);
+		for (std::size_t i=0; i<n; i++)
+		{
+			double a = alpha(i);
+			if (a == 0.0) continue;
+			m_problem.quadratic.row(i, 0, n, q.data());
+			for (std::size_t j=0; j<n; j++) gradient(j) -= a * q[j];
+		}
+		setInitialSolution(alpha, gradient);
+	}
+
 	///\brief Remove the i-th example from the problem.
 	virtual void deactivateVariable(std::size_t i){
 		SIZE_CHECK(i < dimensions());
@@ -370,7 +413,8 @@ public:
 	using base_type::isUpperBound;
 	using base_type::boxMin;
 	using base_type::boxMax;
-		
+	using base_type::setInitialSolution;
+
 	virtual void updateSMO(std::size_t i, std::size_t j){
 		double aiOld = alpha(i);
 		double ajOld = alpha(j);
--- a/include/shark/Algorithms/QP/BoxConstrainedProblems.h
+++ b/include/shark/Algorithms/QP/BoxConstrainedProblems.h
@@ -305,13 +305,13 @@ public:
 		std::size_t n = dimensions();
 		SIZE_CHECK(alpha.size() == n);
 		RealVector gradient = m_problem.linear;
-		std::vector<QpFloatType> q(n);
+		blas::vector<QpFloatType> q(n);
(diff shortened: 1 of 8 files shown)


include/shark/Algorithms/QP/BoxConstrainedProblems.h
```diff
@@ -273,6 +273,49 @@ class BoxConstrainedProblem{
...
+	/// This method can be used to warm-start the solver. It requires a
+	/// feasible solution (alpha), for which it computes the gradient of
+	/// the dual objective function. Note that this is a quadratic time
+	/// operation in the number of non-zero coefficients.
+	void setInitialSolution(RealVector const& alpha)
+	{
+		std::size_t n = dimensions();
+		SIZE_CHECK(alpha.size() == n);
+		RealVector gradient = m_problem.linear;
+		std::vector<QpFloatType> q(n);
```
Ulfgard: RealVector!

include/shark/Algorithms/QP/BoxConstrainedProblems.h
```diff
@@ -273,6 +273,49 @@ class BoxConstrainedProblem{
...
+		for (std::size_t i=0; i<n; i++)
+		{
+			double a = alpha(i);
+			if (a == 0.0) continue;
+			m_problem.quadratic.row(i, 0, n, q.data());
```
Ulfgard: q.storage()

include/shark/Algorithms/QP/BoxConstrainedProblems.h
```diff
@@ -273,6 +273,49 @@ class BoxConstrainedProblem{
...
+			for (std::size_t j=0; j<n; j++) gradient(j) -= a * q[j];
```
Ulfgard: noalias(gradient) -= a*q

include/shark/Algorithms/QP/BoxConstrainedProblems.h
```diff
@@ -370,7 +413,8 @@ struct BoxConstrainedShrinkingProblem : public BoxConstrainedProblem<Problem>{
 	using base_type::isUpperBound;
 	using base_type::boxMin;
 	using base_type::boxMax;
-		
+	using base_type::setInitialSolution;
```
Ulfgard: needed iff you call the function somewhere inside BoxConstrainedShrinkingProblem

include/shark/Algorithms/QP/SvmProblems.h
```diff
@@ -385,7 +385,50 @@ class SvmProblem{
...
+	/// This method can be used to warm-start the solver. It requires a
+	/// feasible solution (alpha), for which it computes the gradient of
+	/// the dual objective function. Note that this is a quadratic time
+	/// operation in the number of non-zero coefficients.
+	void setInitialSolution(RealVector const& alpha)
+	{
+		std::size_t n = dimensions();
+		SIZE_CHECK(alpha.size() == n);
+		RealVector gradient = m_problem.linear;
+		std::vector<QpFloatType> q(n);
```
Ulfgard: same as above.

include/shark/Algorithms/Trainers/CSvmTrainer.h
```diff
@@ -251,6 +266,9 @@ class CSvmTrainer : public AbstractSvmTrainer<
 			typedef SvmShrinkingProblem<SVMProblemType> ProblemType;
 			ProblemType problem(svmProblem,base_type::m_shrinking);
 			QpSolver< ProblemType > solver(problem);
+			if (svm.alpha().size1() > 0 && svm.alpha().size2() == 1) {
+				problem.setInitialSolution(blas::column(svm.alpha(), 0));
```
Ulfgard: i think this will assert if you change C between two runs and the new C is smaller than the old.
also note that the if-statement is always true (setStructure in line 165 will create an alpha matrix with proper size).

Ulfgard:
mostly looks good. but you have to take care that this will never assert. Giving any non-empty model to train should work.

include/shark/Algorithms/QP/BoxConstrainedProblems.h
TGlas: Hmmm... but what's QpFloatType then good for? I could do blas::vector<QpFloatType> though.

include/shark/Algorithms/Trainers/CSvmTrainer.h
TGlas: What you hint at is a serious problem with the current interface. Changing C is relatively easy to deal with. We have to loop over the vector anyway, so checking the constraints is nearly for free. But what if the data set changes, i.e., the basis of the kernel expansion differs from the input portion of the LabeledData object passed to the train function? I could check for equality of the shared pointers inside the data containers - do we have any interface to check whether two data objects share data? We have data.makeIndependent(), but what I need here is data.isSharedWith(otherData). Any objections to adding such a function?

Ulfgard: Hi. You should rather use some kind of equality that checks whether the pointers are the same and in the same order. I think op== does that
________________________________________

TGlas:
Thanks for the comments. I tried to improve things. Please have another look since the logic changed a bit - now the data sets are actually compared.

include/shark/Algorithms/Trainers/CSvmTrainer.h
```diff
@@ -266,9 +266,22 @@ class CSvmTrainer : public AbstractSvmTrainer<
 			typedef SvmShrinkingProblem<SVMProblemType> ProblemType;
 			ProblemType problem(svmProblem,base_type::m_shrinking);
 			QpSolver< ProblemType > solver(problem);
-			if (svm.alpha().size1() > 0 && svm.alpha().size2() == 1) {
-				problem.setInitialSolution(blas::column(svm.alpha(), 0));
+			// truncate the existing solution to the bounds
+			RealVector const& reg = this->regularizationParameters();
+			double C_minus = reg(0);
+			double C_plus = (reg.size() == 1) ? reg(0) : reg(1);
+			for (std::size_t i=0, b=0; b<dataset.labels().numberOfBatches(); b++)
```
Ulfgard: you can make your life easier by iterating over dataset.labels().elements() and thus can skip the for loop over the batches. use autop for the iterator type then, or for(auto label: dataset.labels().elements()) and increment an additional counter.

also: i do not see where you change i in this loop!

TGlas: Damn! You add a unit test and it is still all wrong :)

Ulfgard:
Great. code of the trainer & logic looks okay. now you only need to fix the continuous integration:

https://travis-ci.org/Shark-ML/Shark/jobs/140437943

TGlas:
Weird - everything works locally. And complaining that the definition in dataset.h:150 is redundant to dataset.h:150 is not constructive :) Well, maybe implementing a friend function inline is not the smartest thing to do. Let's see whether the fix makes travis happy.

PR #89, opened 2016-07-02
Data download

TGlas:
Added support for downloading data from online repositories like mldata.org.
The functionality is covered by a unit test and a tutorial.

### DIFF
--- a/Test/Data/Download.cpp
+++ b/Test/Data/Download.cpp
@@ -1,0 +1,29 @@
+#define BOOST_TEST_MODULE ML_Download
+#include <boost/test/unit_test.hpp>
+
+#include <shark/Data/Download.h>
+
+using namespace shark;
+
+
+BOOST_AUTO_TEST_SUITE(Data_Download)
+
+BOOST_AUTO_TEST_CASE(Data_Download_URL)
+{
+	LabeledData<RealVector, unsigned int> dataset;
+	downloadSparseData(dataset, "www.csie.ntu.edu.tw/~cjlin/libsvmtools/datasets/binary/svmguide1");
+	BOOST_CHECK_EQUAL(dataset.numberOfElements(), 3089);
+	BOOST_CHECK_EQUAL(inputDimension(dataset), 4);
+	BOOST_CHECK_EQUAL(numberOfClasses(dataset), 2);
+}
+
+BOOST_AUTO_TEST_CASE(Data_Download_MLData)
+{
+	LabeledData<RealVector, unsigned int> dataset;
+	downloadFromMLData(dataset, "iris");
+	BOOST_CHECK_EQUAL(dataset.numberOfElements(), 150);
+	BOOST_CHECK_EQUAL(inputDimension(dataset), 4);
+	BOOST_CHECK_EQUAL(numberOfClasses(dataset), 3);
+}
+
+BOOST_AUTO_TEST_SUITE_END()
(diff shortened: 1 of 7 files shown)


include/shark/Data/Impl/Downloader.hpp
```diff
@@ -0,0 +1,285 @@
...
+		if (m_handle > 0) ::closesocket(m_handle);
+#else
+		if (m_handle > 0) ::close(m_handle);
+#endif
+		m_handle = 0;
+	}
+
+	/// \brief Read data from the socket (blocking).
+	///
+	/// The operation reads at must #buffersize bytes from the socket
```
Ulfgard: must->most

Test/Data/Download.cpp
```diff
@@ -0,0 +1,29 @@
...
+BOOST_AUTO_TEST_CASE(Data_Download_MLData)
+{
+	LabeledData<RealVector, unsigned int> dataset;
+	downloadFromMLData(dataset, "iris");
+	BOOST_CHECK_EQUAL(dataset.numberOfElements(), 150);
+	BOOST_CHECK_EQUAL(inputDimension(dataset), 4);
+	BOOST_CHECK_EQUAL(numberOfClasses(dataset), 3);
+}
+
+BOOST_AUTO_TEST_SUITE_END()
```
Ulfgard: i like this. the problem is, that this is going to fail if this is build on a system without internet access or with blocked internet access (e.g., for security reasons, unknown build script and all). This will also fail if the server is down. can we add some catches for this case? e.g. check if an exception is thrown that tells us that a connection could not be opened or something or that the server was urneachable? in this case this would not be an error.

include/shark/Data/Impl/Downloader.hpp
```diff
@@ -0,0 +1,285 @@
...
+		std::size_t colon = h.find(":");
+		if (colon == std::string::npos) throw std::runtime_error("[download] http protocol violation");
+		std::string tag = h.substr(0, colon);
+		// convert plain ASCII to lower case
+		for (std::size_t i=0; i<tag.size(); i++) if (tag[i] >= 65 && tag[i] <= 90) tag[i] += 32;
+		std::string value = h.substr(colon + 1);
+		while (! value.empty() && value[0] == ' ') value.erase(0, 1);
+		while (! value.empty() && value[value.size() - 1] == ' ') value.erase(value.size() - 1);
+		// convert plain ASCII to lower case
+		for (std::size_t i=0; i<value.size(); i++) if (value[i] >= 65 && value[i] <= 90) value[i] += 32;
```
Ulfgard: std::tolower

Ulfgard:
looks good. some of the error handling could be better. the socket also needs some documentation for the corner cases as mentioned above, fix and merge :)

include/shark/Data/Impl/Downloader.hpp
TGlas: > std::tolower

Maybe I am overly pessimistic here, but I generally avoid functions that
depend on the "current locale". What if someone compiles this code with,
say, Greek, Arab, or Chinese locale? Will tolower then still do its job?
The problem is that the HTTP standard does not care for a locale.

Test/Data/Download.cpp
TGlas: Do you think that successfully connecting a socket to mldata.org:80 is a sufficient test? It checks for the internet connection on port 80 and for the server being up and running.

PR #102, opened 2016-07-26
Do not include SharkVersion executable to library targets.

ghisvail:
Having the Shark export targets bound to the library and an executable is not beneficial for packaging sake, because libraries and executables are often separated. Unless this was really intended, please consider merging this change which excludes the SharkVersion executable from the library's export.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -58,7 +58,4 @@ install(TARGETS shark
 	    ARCHIVE DESTINATION lib)
 	
 install(TARGETS SharkVersion
-	    EXPORT SharkTargets
 	    RUNTIME DESTINATION bin)
-
-


Ulfgard:
i think this is in accordance to many other libraries that come with an application that just prints configuration information for the maintainers

ghisvail:
> i think this is in accordance to many other libraries that come with an application that just prints configuration information for the maintainers.

Do you have such example in mind? It is the first time I see such setup.

The problem remains that `SharkVersion` becomes part of the exported targets although it contributes nothing to it from a client project perspective. All the client project should care is having the `INCLUDE_DIRS`, `LIBRARY_DIRS` and `LIBRARIES` fields be provided. The library `VERSION` can also be provided with an additional `ConfigVersion` file, which is what CMake recommends in their wiki and what most CMake projects I have worked with do.

Perhaps this PR should be complemented with a redesign of the CMake packaging section in the main CMakeLists.txt. Something that would follow the official CMake [guidelines](https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html). I'll attempt something, but please consider merging this first.

Issue #105, opened 2016-08-10
Segmentation fault when training regressiontree


jakobht:
The program below segfaults due to unbounded calls of CARTTrainer::measureStrength

The program loads input points from the data array declared first, and creates 0 labels for all inputs.

It then tries to learn a regression tree for this data.

The program segfaults in the CARTTrainer::pruneTree method due to unbounded recursive calls of measureStrength (probably because the learned tree has a loop).

I'm not sure if I'm doing something stupid or if it is a bug in the Trainer. 
I have successfully trained other datasets without problems.
The issue is both present in the Release from the website and when cloning from github. 

My platform is Arch linux 
Gcc version 6.1.1
I compile with command:
`g++  -g example.cpp  -lboost_serialization -lshark`

There are 80 points in the data array but I found that it still breaks with 72 points (but not with 71 points).

The program (apparently Github does not allow me to attach a cpp file):

```c++
#include <shark/Algorithms/Trainers/CARTTrainer.h>
#include <iostream>

using namespace std; 
using namespace shark;

double data[] = {
    0,2.49213,2.49213,0,
0,2.95446,2.95446,0,
0,2.77123,2.77123,0,
0,2.11444,2.11444,0,
0,2.63859,2.63859,0,
0,2.19116,2.19116,0,
0,2.0851,2.0851,0,
0,2.13839,2.13839,0,
0,2.73129,2.73129,0,
0,2.21937,2.21937,0,
0,2.10212,2.10212,0,
0,2.74708,2.74708,0,
0,2.81297,2.81297,0,
0,2.55067,2.55067,0,
0,2.82431,2.82431,0,
0,2.88559,2.88559,0,
0,2.21894,2.21894,0,
0,2.80792,2.80792,0,
0,2.23107,2.23107,0,
0,2.1189,2.1189,0,
0,2.40597,2.40597,0,
0,2.02819,2.02819,0,
0,2.93205,2.93205,0,
0,2.02587,2.02587,0,
0,2.23181,2.23181,0,
0,2.02214,2.02214,0,
0,2.3436,2.3436,0,
0,2.51296,2.51296,0,
0,2.27819,2.27819,0,
0,2.04024,2.04024,0,
0,2.00951,2.00951,0,
0,2.18167,2.18167,0,
0,2.15445,2.15445,0,
0,2.8293,2.8293,0,
0,2.35753,2.35753,0,
0,2.55998,2.55998,0,
0,2.50951,2.50951,0,
[...]

PR #107, opened 2016-08-13
The nodeId assignment in the CART classification and regression are now the same

jakobht:
This fixes an overflow issue when generating the right branch of the
trees.

The problem was that when all the labels are 0 only the right branch of the tree will be generated, that meant that the ids became very large very fast. This meant that there was an overflow in the ids, leading to a cycle in the tree. 

When the tree walk was then attempted it would recur forever.

fixes Shark-ML/Shark#105


Ulfgard:
good catch! thank you!

PR #108, opened 2016-08-16
Fixed the issue that one node trees are not set in CART

jakobht:
Moved the stopping condition of the loop in the CARTTrainer to accound
for trees with only one node.


Ulfgard:
just to understand: the issue here is that we might prune all nodes in a tree? would this then loop forever?

jakobht:
The issue is that if the buildTree method only gives trees which have excatly one node, then the condition in the while was never satisfied, that meant that bestTree was never set to anything but an empty vector.

This would then mean a segfault on line 348 in CARTClassifier when trying to do lookups (as the m_tree in the CARTClassifier was empty).

PR #109, opened 2016-08-25
Fix HDF5 detection with CMake 3.6.

ghisvail:
This patch fixes the HDF5 setup with the latest CMake 3.6.

`HDF5_INCLUDE_DIR` has been  deprecated anyway (and is currently broken), so it is replaced by `HDF5_INCLUDE_DIRS`.

`HDF5_LIBRARY_DIR` is not necessary (and no longer set), since `HDF5_LIBRARIES` contains full paths to the libraries.

Finally, CMake honors the separation between the low-level library (`HDF5_LIBRARIES`) and high-level bindings (`HDF5_HL_LIBRARIES`), so both must be provided to `LINK_LIBRARIES`.


Ulfgard:
thank you!

PR #117, opened 2016-09-27
Random forests optimization

wrigleyster:
Includes a number of significant speedups, seeding of random functions, and now properly calculated OOB estimates.

Still in the merging pipe line:
* All equivalent CART improvements
* Choice of bootstrapping method
* Choice of impurity measure for Classification
* Some cleanup involving the old method for OOB estimates

### DIFF
--- a/include/shark/Algorithms/Trainers/CARTTrainer.h
+++ b/include/shark/Algorithms/Trainers/CARTTrainer.h
@@ -110,6 +110,8 @@ protected:
 
 	typedef ModelType::TreeType TreeType;
 
+	// LabelVector
+	using ClassVector = UIntVector;
 
 	///Number of attributes in the dataset
 	std::size_t m_inputDimension;
@@ -129,14 +131,14 @@ protected:
 	//Classification functions
 	///Builds a single decision tree from a classification dataset
 	///The method requires the attribute tables,
-	SHARK_EXPORT_SYMBOL TreeType buildTree(AttributeTables const& tables, ClassificationDataset const& dataset, boost::unordered_map<std::size_t, std::size_t>& cAbove, std::size_t nodeId );
+	SHARK_EXPORT_SYMBOL TreeType buildTree(AttributeTables const& tables, ClassificationDataset const& dataset, ClassVector& cAbove, std::size_t nodeId );
 
 	///Calculates the Gini impurity of a node. The impurity is defined as
 	///1-sum_j p(j|t)^2
 	///i.e the 1 minus the sum of the squared probability of observing class j in node t
-	SHARK_EXPORT_SYMBOL double gini(boost::unordered_map<std::size_t, std::size_t>& countMatrix, std::size_t n);
-	///Creates a histogram from the count matrix.
-	SHARK_EXPORT_SYMBOL RealVector hist(boost::unordered_map<std::size_t, std::size_t> countMatrix);
+	SHARK_EXPORT_SYMBOL double gini(ClassVector const& countVector, std::size_t n) const;
+	///Creates a histogram from the count vector.
+	SHARK_EXPORT_SYMBOL RealVector hist(ClassVector const& countVector) const;
 
 	///Regression functions
 	SHARK_EXPORT_SYMBOL TreeType buildTree(AttributeTables const& tables, RegressionDataset const& dataset, std::vector<RealVector> const& labels, std::size_t nodeId, std::size_t trainSize);
@@ -162,7 +164,7 @@ protected:
 	///Splits the attribute tables by a attribute index and value. Returns a left and a right attribute table in the variables LAttributeTables and RAttributeTables
 	SHARK_EXPORT_SYMBOL void splitAttributeTables(AttributeTables const& tables, std::size_t index, std::size_t valIndex, AttributeTables& LAttributeTables, AttributeTables& RAttributeTables);
 	///Crates count matrices from a classification dataset
-	SHARK_EXPORT_SYMBOL boost::unordered_map<std::size_t, std::size_t> createCountMatrix(ClassificationDataset const& dataset);
+	SHARK_EXPORT_SYMBOL CARTTrainer::ClassVector createCountVector(ClassificationDataset const& dataset) const;
 
 
 };
--- a/include/shark/Algorithms/Trainers/CARTTrainer.h
+++ b/include/shark/Algorithms/Trainers/CARTTrainer.h
@@ -116,14 +116,13 @@ protected:
 	///Number of attributes in the dataset
 	std::size_t m_inputDimension;
 
-	///Size of labels
+	///Dimension of a label. Used in Regression
 	std::size_t m_labelDimension;
+	///Holds the number of distinct labels. Used in Classification
+	std::size_t m_labelCardinality;
 
 	///Controls the number of samples in the terminal nodes
 	std::size_t m_nodeSize;
-
-	///Holds the maximum label. Used in allocating the histograms
-	unsigned int m_maxLabel;
 
 	///Number of folds used to create the tree.
 	unsigned int m_numberOfFolds;
--- a/include/shark/Algorithms/Trainers/CARTTrainer.h
+++ b/include/shark/Algorithms/Trainers/CARTTrainer.h
@@ -112,6 +112,7 @@ protected:
 
 	// LabelVector
 	using ClassVector = UIntVector;
+	using LabelVector = std::vector<RealVector>;
 
 	///Number of attributes in the dataset
 	std::size_t m_inputDimension;
--- a/include/shark/Algorithms/Trainers/CARTcommon.h
+++ b/include/shark/Algorithms/Trainers/CARTcommon.h
@@ -1,0 +1,156 @@
+//===========================================================================
+/*!
+ *
+ *
+ * \brief       General functions for Tree modeling.
+ *
+ *
+ *
+ * \author      K. N. Hansen, J. Kremer, J. Wrigley
+ * \date        2011-2016
+ *
+ *
+ * \par Copyright 1995-2015 Shark Development Team
+ *
+ * <BR><HR>
+ * This file is part of Shark.
+ * <http://image.diku.dk/shark/>
+ *
+ * Shark is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Shark is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Shark.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+//===========================================================================
+#ifndef SHARK_ALGORITHMS_TRAINERS_CARTCOMMON_H
+#define SHARK_ALGORITHMS_TRAINERS_CARTCOMMON_H
+
+#include <shark/Models/Trees/General.h>
+#include <shark/Data/DataView.h>
+#include <vector>
+#include <unordered_map>
+#include <utility>
+namespace shark {
+namespace detail {
+namespace cart {
+/// ClassVector
+using ClassVector = UIntVector;
+
+
+struct Attribute {
+	double value;
+	std::size_t id;
+
+	inline bool operator<(Attribute const &v) const {
+		return value < v.value;
+	}
+};
+
+
+class Index {
+/// attribute table
+	using AttributeTable = std::vector<Attribute>;
+/// collecting of attribute tables
+	using AttributeTables = std::vector<AttributeTable>;
+
+	std::size_t m_noInputDimensions, m_noElements;
+	AttributeTables m_tables;
+
+	explicit Index(AttributeTables &&tables)
+			: m_noElements(tables[0].size()),
+			  m_noInputDimensions(tables.size()),
+			  m_tables(std::move(tables))
+	{}
+public:
+	struct size_pair{
+		std::size_t n_inputDimensions, n_elements;
+		size_pair(std::size_t n_inputDimensions, std::size_t n_elements)
+				: n_inputDimensions(n_inputDimensions),
+				  n_elements(n_elements)
+		{}
+	};
+
+/** Creates an index of the dataset
+ *  A dataset consisting of m input variables has m sorted attribute tables.
+ *  [attribute | rid ]
+ */
+	template<class Dataset>
+	explicit Index(DataView<Dataset const> const &elements)
+			: m_noElements{elements.size()},
+			  m_noInputDimensions{elements[0].input.size()},
+			  m_tables(m_noInputDimensions)
+	{
+		std::size_t n_elements = m_noElements;
+		//Each entry in the outer vector is an attribute table
+		//For each column
+		for (std::size_t j = 0; j < m_noInputDimensions; j++) {
+			//m_tables.push_back({});
+			auto &table = m_tables[j];
+			table.reserve(n_elements);
+
+			//For each row, store Attribute value, class and rowId
+			for (std::size_t i = 0; i < n_elements; i++) {
+				table.push_back(Attribute{elements[i].input[j], i});
+			}
+			std::sort(table.begin(), table.end());
+		}
+	}
+
+/**
+ * Returns two Indices: left and right
+ * Calculated from splitting tables at (index, valIndex)
+ */
+	std::pair<Index, Index> split(std::size_t index, std::size_t valIndex) {
+		//Build a hash table for fast lookup
+		std::unordered_map<std::size_t, bool> hash;
+		for(std::size_t i = 0, s = m_tables[index].size(); i<s; ++i) {
+			hash[m_tables[index][i].id] = (i<=valIndex);
+		}
+
+		AttributeTables RAttributeTables, LAttributeTables;
+		for(auto && table : m_tables) {
+			auto begin = table.begin(), end = table.end();
+			auto middle = std::stable_partition(begin,end,[&hash](Attribute const& entry){
+				return hash[entry.id];
+			});
+			RAttributeTables.emplace_back(AttributeTable{middle,end});
+			table.resize(std::distance(begin,middle));
+			LAttributeTables.emplace_back(std::move(table));
+		}
+		m_tables.clear(); m_tables.shrink_to_fit();
+		return std::make_pair(Index{std::move(LAttributeTables)},
+							  Index{std::move(RAttributeTables)});
+	}
+
+	inline size_pair size() const {
+		return {m_noInputDimensions,m_noElements};
+	}
+	inline AttributeTable const& operator[](std::size_t i) const {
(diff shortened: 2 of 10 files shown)


Ulfgard:
Hi,

thanks. Note that the continuous integration is failing. VC does not yet support defaulting the move constructor/assignment operator and you have a missing include for clang.

wrigleyster:
Ah, right, thank you :) sorry, I was waiting for it to finish, and then I forgot all about it. I'll quickly fix it up ;)

Sent from my phone

src/Algorithms/CARTTrainer.cpp
```diff
@@ -529,12 +527,10 @@ void CARTTrainer::splitAttributeTables(AttributeTables const& tables, std::size_
 ///Calculates the Gini impurity of a node. The impurity is defined as
 ///1-sum_j p(j|t)^2
 ///i.e the 1 minus the sum of the squared probability of observing class j in node t
-double CARTTrainer::gini(boost::unordered_map<std::size_t, std::size_t>& countMatrix, std::size_t n){
+double CARTTrainer::gini(ClassVector const& countVector, std::size_t n) const {
 	double res = 0;
```
Ulfgard: res = sum(sqr(countVector/static_cast<double>(n))); //(hopefully, if i did not fuck up in the linear algebra...)

also din't you define a free function gini as well?

include/shark/Algorithms/Trainers/CARTTrainer.h
```diff
@@ -110,33 +110,35 @@ class CARTTrainer
...
 
 	///Calculates the Gini impurity of a node. The impurity is defined as
 	///1-sum_j p(j|t)^2
 	///i.e the 1 minus the sum of the squared probability of observing class j in node t
-	SHARK_EXPORT_SYMBOL double gini(boost::unordered_map<std::size_t, std::size_t>& countMatrix, std::size_t n);
-	///Creates a histogram from the count matrix.
-	SHARK_EXPORT_SYMBOL RealVector hist(boost::unordered_map<std::size_t, std::size_t> countMatrix);
+	SHARK_EXPORT_SYMBOL double gini(ClassVector const& countVector, std::size_t n) const;
+	///Creates a histogram from the count vector.
+	SHARK_EXPORT_SYMBOL RealVector hist(ClassVector const& countVector) const;
```
Ulfgard: didn't you add a free function for that?)

src/Algorithms/CARTcommon.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include <shark/Algorithms/Trainers/CARTcommon.h>
+#include <shark/Core/Math.h>
+#include <shark/Algorithms/Trainers/RFTrainer.h>
+
+namespace shark {
+namespace detail{
+namespace cart {
+
+RealVector hist(ClassVector const& countVector)
+{
```
Ulfgard: RealVector normedHistogram = countVector/double(sum(countVector));
return normedHistogram;

src/Algorithms/CARTTrainer.cpp
```diff
@@ -245,42 +244,43 @@ void CARTTrainer::measureStrength(TreeType & tree, std::size_t nodeId, std::size
...
+				cBestAbove(m_labelCardinality);//labels of best split
 
-		for (std::size_t attributeIndex=0; attributeIndex < m_inputDimension; attributeIndex++){
+		for (std::size_t attributeIndex=0; attributeIndex < m_inputDimension; ++attributeIndex){
 			AttributeTable const& table = tables[attributeIndex];
-			boost::unordered_map<std::size_t, std::size_t> cTmpAbove = cAbove;
-			boost::unordered_map<std::size_t, std::size_t> cBelow;
-			for(std::size_t i=0; i<n-1; i++){//go through all possible splits
+			auto cTmpAbove = cAbove;
+			std::fill(cBelow.begin(),cBelow.end(),0);
```
Ulfgard: cBelow.clear();

src/Algorithms/CARTcommon.cpp
```diff
@@ -0,0 +1,108 @@
...
+///1-sum_j p(j|t)^2
+///i.e the 1 minus the sum of the squared probability of observing class j in node t
+double gini(ClassVector const& countVector, std::size_t n)
+{
+	if(!n) return 1;
+
+	double res = 0.;
+	n *= n;
+	for(auto const& i: countVector){
+		res += sqr(i)/ static_cast<double>(n);
```
Ulfgard: in your implementation above you used res += sqr(i/ static_cast<double>(n)); which one is correct?
also again: sum(sqr(countVector))/static_cast<double(n);

src/Algorithms/CARTcommon.cpp
```diff
@@ -0,0 +1,108 @@
...
+	}
+	return 1-res;
+}
+
+// ME = 1-max(count)/n
+double misclassificationError(ClassVector const& countVector, std::size_t n)
+{
+	if(!n) return 1.;
+	auto m = countVector[0];
```
Ulfgard: auto m = max(countVector);
better double, because see below.

src/Algorithms/CARTcommon.cpp
```diff
@@ -0,0 +1,108 @@
...
+	for(auto c: countVector){
+		m = std::max(m,c);
+	}
+	return 1-m/n;
```
Ulfgard: m and n are integers and m<n. Therefore the result is going to be 0.

src/Algorithms/CARTcommon.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// CE = - sum_j(count[j] log(count[j]/n))/n
+double crossEntropy(ClassVector const& countVector, std::size_t n)
+{
+	return - sum<double>(0,n,[&](std::size_t i){
+		if(!countVector[i]) return 0.;
+		return countVector[i]*std::log(countVector[i]/n);
```
Ulfgard: you are again dividing integers.

src/Algorithms/RFTrainer.cpp
```diff
@@ -238,385 +241,181 @@ void RFTrainer::train(RFClassifier& model, ClassificationDataset const& dataset)
...
+		// this shouldn't really happen: leaves ought to be cought by previous check
+		// TODO(jwrigley): Why is hist only applied to leaves, when average is applied to all nodes
+		nodeInfo.label = detail::cart::hist(cFull);
 		return tree;
 	}
+	nodeInfo <<= split;
+	auto lrTables = tables.split(split.splitAttribute, split.splitRow);
+	//Continue recursively
+
+	nodeInfo.leftNodeId  = (nodeId<<1)+1;   nodeInfo.rightNodeId = (nodeId<<1)+2;
```
Ulfgard: this is going to fail with trees that for some reason are very deep. We had this in the CARTTrainer. the logic should be after building the tree:

nodeInfo.leftNodeId = nodeId+1;
...
nodeInfo.rightNodeId = nodeInfo.leftNodeId + lTree.size();

This also happens in the other case

src/Algorithms/RFTrainer.cpp
```diff
@@ -238,385 +241,181 @@ void RFTrainer::train(RFClassifier& model, ClassificationDataset const& dataset)
...
-	}
+	//Split the attribute tables
+	auto lrTables = tables.split(split.splitAttribute, split.splitRow);
 
-	if(isLeaf){
-		tree.push_back(nodeInfo);
-		return tree;
-	}
+	//Continue recursively
+	nodeInfo.leftNodeId = (nodeId<<1)+1;
```
Ulfgard: same index problem here

include/shark/Algorithms/Trainers/CARTcommon.h
```diff
@@ -0,0 +1,156 @@
...
+			RAttributeTables.emplace_back(AttributeTable{middle,end});
+			table.resize(std::distance(begin,middle));
+			LAttributeTables.emplace_back(std::move(table));
+		}
+		m_tables.clear(); m_tables.shrink_to_fit();
+		return std::make_pair(Index{std::move(LAttributeTables)},
+							  Index{std::move(RAttributeTables)});
+	}
+
+	inline size_pair size() const {
```
Ulfgard: i would use two different functions here. size() has a very special name in c++, not to tinker with :)

include/shark/Models/Trees/General.h
```diff
@@ -0,0 +1,126 @@
...
+#ifndef SHARK_MODELS_TREES_GENERAL_H
+#define SHARK_MODELS_TREES_GENERAL_H
+#include <memory>
+template<class T> class TD;
+#define typeof(param) TD<decltype(param)>{}
+namespace shark {
+namespace detail {
+namespace cart {
+// Helper functions
+template<class T> using sink = T;
```
Ulfgard: what is the purpose of sink?

include/shark/Models/Trees/CARTClassifier.h
```diff
@@ -86,6 +86,15 @@ class CARTClassifier : public AbstractModel<RealVector,LabelType>
 			ar & r;
 			ar & g;
 		}
+		NodeInfo() {}
+		explicit NodeInfo(std::size_t nodeId) : nodeId(nodeId) {}
+		NodeInfo(std::size_t nodeId, LabelType const& label) : nodeId(nodeId), label(label) {}
+		NodeInfo(std::size_t nodeId, LabelType && label)
+				: nodeId(nodeId), label(std::move(label)) {}
+		NodeInfo(NodeInfo const&) = default;
+		NodeInfo(NodeInfo &&) = default;
```
wrigleyster: They won't be auto-generated so long as there are any declared constructors, however, you might still right that we don't need the two of them as `label' is likely small (<10 elements), and thus would not benefit from the move, if blas::vector implements the small-vector optimization. 

If there is no smal-vector optimization, I can also rewrite them using boost macros, so we still get the move support on supporting compilers?

include/shark/Models/Trees/General.h
wrigleyster: When compiled `sink<T&>` evaluates to `T&`. `sink<T&>` is used to mark parameters that will be moved from eventually, and whose state therefore will be undefined subsequently.

It solves 2 problems:
1. It informs the user that we intend to invalidate the object later by moving from it. The alternative way to do this would be to make the function take a `T&&` and call it `func(std::move(arg))`, but this doesn't solve the second problem.
2. In cases when move can't actually be performed (say the object being moved doesn't implement move semantics (yet)), it is faster to use `sink<T&>` going down the call-chain and then `std::move`, rather than to `std::move` on every call down, because in those cases move would then become copy.

sink & move thus evaluates to reference & copy, rather than n*copy in situations when move-semantics aren't fully supported.

include/shark/Models/Trees/General.h
```diff
@@ -0,0 +1,126 @@
...
+#define CART_STRINGIFY(num) CART_STRINGIFY_DETAIL(num)
+#define CART_Expects(cond) \
+	if(!cond){ \
+		throw shark::detail::cart::fail_fast{"expectation failed at " \
+        __FILE__ ": " CART_STRINGIFY(__LINE__)}; \
+	}
+
+#define DEBUG(msg) std::cout << msg << std::endl
+template<class Iterator>
+inline std::size_t argmax(Iterator begin, Iterator end) {
```
wrigleyster: brilliant! I'll just use that :)

include/shark/Models/Trees/General.h
```diff
@@ -0,0 +1,126 @@
...
+}
+template<class Container, class F>
+inline Container generate_i(std::size_t const length, F&& f){
+	Container v(length);
+	for(std::size_t i=0; i<length; ++i) v[i] = f(i);
+	return v;
+}
+
+template<class T, class F>
+inline T stable_sum(std::size_t i, std::size_t n, F&& f) {
```
wrigleyster: Indeed you are right. There is only need for one sum-function, I just couldn't make up my mind which one. The two functions both produce different results on moderately large floating-point input. I named the "stable" one thus, because as best I could tell it would be the more accurate one due smaller and fewer rounding errors (because it adds like numbers). The `stable_sum` *is* slower by a small margin however.

Whether we hit the 8 digit difference depends on the dataset size. I'd guess we hit the limit of the stack before that though..

src/Algorithms/RFTrainer.cpp
wrigleyster: Ah, right, for balanced trees the two would be equivalent, but for unbalanced trees we end up reserving too much of the id-space for unused branches, resulting in overflow. Gotcha! Fix coming up ;)

src/Algorithms/CARTTrainer.cpp
wrigleyster: Unfortunately `std::fill` can't be replaced by `cBelow.clear`, because `clear` leaves the vector empty, when we need it to still have its length, just zeroed.

That said, the changes to CARTTrainer.* in this changeset only include the unordered_map → vector conversion. The next pull-request will include many of the same changes as done to the RFTrainers. After that the splitting will occur in much the same way as in RF, and we'll be done with `std::fill` :)

src/Algorithms/CARTTrainer.cpp
```diff
@@ -310,37 +310,36 @@ CARTTrainer::TreeType CARTTrainer::buildTree(AttributeTables const& tables, Clas
...
 	}
-	
+
 	TreeType tree;
 	tree.push_back(nodeInfo);
 	return tree;
 }
 
-RealVector CARTTrainer::hist(boost::unordered_map<std::size_t, std::size_t> countMatrix){
+RealVector CARTTrainer::hist(ClassVector const& countVector) const {
```
wrigleyster: Wow, thank you, I had completely forgotten about that! Initially `countVector` was a `std::vector`, I changed it to a UIntVector specifically for this reason, but then just completely forgotten about it. I'll put the impurity functions through the same magic :D

src/Algorithms/CARTcommon.cpp
wrigleyster: uh, thanks! well spotted.

src/Algorithms/CARTcommon.cpp
wrigleyster: thanks!

include/shark/Algorithms/Trainers/CARTcommon.h
wrigleyster: Good point, will do.

include/shark/Algorithms/Trainers/CARTcommon.h
```diff
@@ -0,0 +1,156 @@
...
+	double value;
+	std::size_t id;
+
+	inline bool operator<(Attribute const &v) const {
+		return value < v.value;
+	}
+};
+
+
+class Index {
```
wrigleyster: Honestly I was struggling a little with the naming of this class. (Also with `AttributeTable`(`s`) I feel there are more sensical names out there, I just couldn't think of them. Suggestions are welcome.)

`Index` groups the functionality of the previous
`createAttributeTables()`, and
`splitAttributeTables()`
functions. Basically an `Index` indexes the dataset by attribute, providing fast sorted-order lookup of each attribute. The use case is, that we need to scan attributes in order of size when looking for a split.

How about this(below), would this be an adequate description?
```diff
+/**
+ * An index of a dataset providing fast sorted-order lookup of each attribute.
+ */
 class Index {
```

Perhaps a better name for the class would be `SortedIndex`?

include/shark/Models/Trees/CARTClassifier.h
Ulfgard: hi the issue is only that you can not write "= default". You can still implement it by yourself (by using the naive implementation).

include/shark/Models/Trees/General.h
Ulfgard: I like the T&& solution better. As we are writing our own code here (mostly) or rely on boost we can assume that we can always implement the move operators. Right now it is more important that we are consistent across shark. so either we have a global shark wide sink (which than had to be used) or none at all.

include/shark/Models/Trees/General.h
Ulfgard: we should use a normal sum than. it is not so important that we have high precision as we would assume that any decision based on roundoff error is pretty much random from the start. (most machine learning libraries favour float instead of double as default because of this reason)

src/Algorithms/CARTTrainer.cpp
Ulfgard: its a blas::vector when i remember correctly. that just zeroes its content (the semantiv of vectors in linear algebra are unfortunately a bit different than the misnamed std::vector). anyway if it is removed at some point no need for argue then :)

src/Algorithms/CARTcommon.cpp
wrigleyster: They're both correct, this one merely squares `n` in advance (line 66)

src/Algorithms/CARTcommon.cpp
```diff
@@ -35,63 +35,40 @@
...
 
-// CE = - sum_j(count[j] log(count[j]/n))/n
+// CE = - sum_j(count[j]/n * log(count[j]/n))
 double crossEntropy(ClassVector const& countVector, std::size_t n)
 {
-	return - sum<double>(0,n,[&](std::size_t i){
-		if(!countVector[i]) return 0.;
-		return countVector[i]*std::log(countVector[i]/n);
-	})/n;
+	if(!n) return std::numeric_limits<double>::infinity();
```
Ulfgard: hi,

if p->0 then p*log(p)->0, not infinity.

moreover the lines below will give you trouble if p is 0 for any class as 0*log(0) is problematic (but should be 0). This is a frequent case in multi-class problems in which a split might easily lead to removal of a class (e.g. number of points in the node < number of classes). The right computation is summing over the nonzero elements.

include/shark/Models/Trees/RFClassifier.h
```diff
@@ -72,7 +72,7 @@ class RFClassifier : public MeanModel<CARTClassifier<RealVector> >
 		auto n_elements = elements.size();
 		m_OOBerror = detail::cart::sum<double>(n_elements, [&](size_t i){
 			auto y = elements[i].label;
-			auto z = detail::cart::argmax(row(oobClassTally,i));
+			auto z = blas::arg_max(row(oobClassTally,i));
```
Ulfgard: ADL will find the function even without blas:: (same as with row(...))

src/Algorithms/CARTcommon.cpp
wrigleyster: Odd, I was sure I tested this and got a similar (ie. only slightly different) result back as previously with `crossEntropy` in the end I figured `blas::log` was doing something funky to avoid all the infinities, and the fact I was getting numbers back at all threw me off the scent, but you're absolutely right, we need to filter out the zeroes.

As for the return of infinity.. I'm testing `n` not `p`, so technically as `n` → +0,  `p` → infinity and `-p*log(p)` → &minus;infinity, meaning I ought to return the negative infinity. Admittedly though, I was coming at it from a somewhat more brutish point of view: We're measuring impurity, high impurity is bad, if `n==0` that is also bad, thus return a disincentive.. and speaking of... if `n` → &minus;0 then `-p*log(p)` → +infinity, so maybe it's kind of alright?

wrigleyster:
The Travis CI Apple build says: 'cmake command not found'... ?

Ulfgard:
maybe they changed the architecture again. i think we need 

then brew update && brew unlink cmake && brew install cmake && brew link cmake; fi

in the .travis.yml

PR #138, opened 2017-01-17
Added noisy function handling to the CMA-ES

Ulfgard:
Added noisy function handling to CMA-ES

this change includes two parts:
1. Adding a trait to AbstractObjectiveFunction to mark a noisy function
2. Add Hansens noise handling algorithm with slight changes to the CMA-ES

### DIFF
--- a/Test/Algorithms/DirectSearch/CMA.cpp
+++ b/Test/Algorithms/DirectSearch/CMA.cpp
@@ -14,6 +14,51 @@
 
 using namespace shark;
 
+
+struct MultiplicativeNoisySphere : public SingleObjectiveFunction {
+	
+	MultiplicativeNoisySphere(std::size_t numberOfVariables, double sigma):m_numberOfVariables(numberOfVariables), m_sigma(sigma) {
+		m_features |= CAN_PROPOSE_STARTING_POINT;
+		m_features |= IS_NOISY;
+	}
+
+	/// \brief From INameable: return the class name.
+	std::string name() const
+	{ return "MultiplicativeNoisySphere"; }
+
+	std::size_t numberOfVariables()const{
+		return m_numberOfVariables;
+	}
+	
+	bool hasScalableDimensionality()const{
+		return true;
+	}
+
+	void setNumberOfVariables( std::size_t numberOfVariables ){
+		m_numberOfVariables = numberOfVariables;
+	}
+
+	SearchPointType proposeStartingPoint() const {
+		RealVector x(numberOfVariables());
+
+		for (std::size_t i = 0; i < x.size(); i++) {
+			x(i) = Rng::gauss(0,1);
+		}
+		return x;
+	}
+
+	double eval(SearchPointType const& x) const {
+		SIZE_CHECK(x.size() == numberOfVariables());
+		m_evaluationCounter++;
+		double f = norm_sqr(x);
+		double noise = exp(Rng::gauss(0,m_sigma));
+		return f * noise;
+	}
+private:
+	std::size_t m_numberOfVariables;
+	double m_sigma;
+};
+
 BOOST_AUTO_TEST_SUITE (Algorithms_DirectSearch_CMA)
 
 BOOST_AUTO_TEST_CASE( CMA_Cigar )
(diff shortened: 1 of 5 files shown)


Test/Algorithms/DirectSearch/CMA.cpp
```diff
@@ -14,6 +14,51 @@
...
+			x(i) = Rng::gauss(0,1);
+		}
+		return x;
+	}
+
+	double eval(SearchPointType const& x) const {
+		SIZE_CHECK(x.size() == numberOfVariables());
+		m_evaluationCounter++;
+		double f = norm_sqr(x);
+		double noise = exp(Rng::gauss(0,m_sigma));
```
TGlas: I like the log-normal approach, but the standard thing in analysis papers is something like
f *= (1 + sigma * gauss)
or
f *= (1 + sigma * uniform)
That's actually very different since noisy values can become negative in some cases, which completely destroys elitist methods (well, usually step size adaptation kills the method already at much smaller noise levels).

src/Algorithms/DirectSearch/CMA.cpp
```diff
@@ -37,9 +37,21 @@
...
 using namespace shark;
 
+namespace{
+	//computes percentile({|1-R|,|2-R|,...,|N-R|},q)
+	int deltaLim(int rank, std::size_t N, double percentile){
+		blas::vector<int> rankDistr(N);
+		std::iota(rankDistr.begin(),rankDistr.end(),1-rank);
+		noalias(rankDistr) = abs(rankDistr);
+		std::sort(rankDistr.begin(),rankDistr.end());
+		return rankDistr[std::size_t(percentile * (rankDistr.size()-1))];
```
TGlas: Maybe consider using std::nth_element instead to reduce the complexity?
Well, I know, this code is not a performance bottleneck.

src/Algorithms/DirectSearch/CMA.cpp
```diff
@@ -312,7 +338,89 @@ void CMA::updatePopulation( std::vector<IndividualType> const& offspring ) {
...
+			
+			//compute and save new function values of the population
+			penalizingEvaluator(function,offspring.begin(),offspring.begin() + lambdaReeval);
+			std::vector<shark::KeyValuePair<double,std::size_t> > ranksNew(m_lambda);
+			for(std::size_t i = 0; i != m_lambda; ++i){
+				ranksNew[i].key = offspring[i].penalizedFitness();
+				ranksNew[i].value = i;
+			}
+			
+			// update function values of the population to be the mean of old and new values,
```
TGlas: This comment should probably contain a reference to the paper. The reference could also go into the class doxygen header.

src/Algorithms/DirectSearch/CMA.cpp
```diff
@@ -312,7 +338,89 @@ void CMA::updatePopulation( std::vector<IndividualType> const& offspring ) {
...
+				//minus a percentile of the expected difference on a truely random function
+				//as the first and second rank are interchangeable, the average is computed
+				s -= 0.5 * deltaLim(ranks[i].second - (ranks[i].second > ranks[i].first), 2*m_lambda-1,m_rankChangeQuantile);
+				s -= 0.5 * deltaLim(ranks[i].first - (ranks[i].first > ranks[i].second), 2*m_lambda-1,m_rankChangeQuantile);
+			}		
+			s /= lambdaReeval;
+			//simple adaptation of the number of reevaluations
+			if(s > 0){
+				double rawIncrease = (m_numEvalIncreaseFactor - 1.0 )* m_numEvaluations;
+				m_numEvaluations += std::max<std::size_t>(1,std::size_t(rawIncrease+0.5));
```
TGlas: size_t(std::round(rawIncrease)) instead of size_t(rawIncrease + 0.5) to clarify the intent?
Really just a detail.
Maybe a more important note: I don't see immediately why the RHS of += cannot be zero; that would kill the method. Maybe add a std::max(1.0, ...) somewhere.

src/Algorithms/DirectSearch/CMA.cpp
```diff
@@ -312,7 +338,89 @@ void CMA::updatePopulation( std::vector<IndividualType> const& offspring ) {
...
+			}else if (s < 0 && m_numEvaluations > 1){
+				double rawDecrease = (1.0 - 1.0/m_numEvalIncreaseFactor) * m_numEvaluations;
+				m_numEvaluations -= std::max<std::size_t>(1,std::size_t(rawDecrease+0.5));
```
TGlas: Same here, std::round and can the decrease be zero? Of course, m_numEvaluations must be lower bounded by one, and it must be protected against overflow.

Test/Algorithms/DirectSearch/CMA.cpp
Ulfgard: I have changed this and somehow understand why you would like this, as this will make the worst points of the set appear the best when a negative epsilon is drawn. However, i would argue that this test is actually a lot easiuer than the one i came up with: simply because this test is likely to generate more extreme reorderings by single "bad" draws (e.g. worst->best) which the current mechanism is much better in picking up. The result is that the noise estimate is much more stable with your proposed distribution, even though the noise level is much smaller. I have still changed to the first version you proposed.

src/Algorithms/DirectSearch/CMA.cpp
Ulfgard: done

src/Algorithms/DirectSearch/CMA.cpp
Ulfgard: it already has a std::max(1,...) in the RHS

i will use std::lround

src/Algorithms/DirectSearch/CMA.cpp
Ulfgard: raw decrease can be smaller than 0.5 for small number of evaluations (e.g. case = 2). for this case there is already an std::max on the RHS of -=. Note that the LHS is guarantueed to be >=0 for (1.0 - 1.0/m_numEvalIncreaseFactor) < 0.5

PR #147, opened 2017-02-13
Batch interface

### LINKED ISSUE #17: c++11 incompatibility between shark::size and std::size
from the mailing list

"Compiling shark with VS14 and boost 1.59 I get an error "'shark::size' ambiguous call to overloaded function". The ambiguity is between std::size and shark::size, and in the affected files both "using namespace std" and "using namespace shark" are defined. Adding namespace shark to size() solves the problem, but it would be nice to not have to manually do this. "

shark::size got introduced because at that date, std::size was not available. As shark does not officially support C++11 yet, we can defer the fix until its time

Ulfgard:
Rather large rework of the Batch Interface. The goal was to make it slightly easier to maintain, while fixing one or two issues.

Changes:
- fixes #17 
- reworked the ranges and iterator interface of the Datasets.
  Basically, we implement the batch(i) method and a way to get the jth element of a batch. Using this we can then iterate over the element using indices i and j. This way, implementing iterators and ranges over the elements is much more straight forward. similarly, iterating over the batches is done using an iterator going over the indices. Everything is a bit more complicated for LabeledData as the batches returned are purely virtual beasts. So there is some amount of template hackery involved to get that to work out nicely.
- Previously batches where implemented using boot::range, but this clashes a lot with how boost::range is used. So we now have an interface for the most common batch methods:
  * getBatchElement(batch,j) returns the jth element of the batch
  * batchSize(batch) returns the number of elements in the batch
  * batchBegin(batch) returns an iterator to the beginning of the batch
  * batchEnd(batch) returns an iterator to the end of the batch
  * Everything is strung together using two main traits classes: Batch\<T> tells the system how batches holding elements of type T work. this is required to get the mapping batches<->matrices. BatchTraits\<BT> then tells the system which Batch\<T> is responsible for a batch BT (simplest case: BatchTraits\<BT> -> Batch\<BT::value_type> ). 

There are many weird special cases, especially considering proxy classes that behave like Batch<T> but are not. This makes everything a bit unstable, e.g. we have to be vary careful that the right swap is used (not std::swap!) and that proxies are not accidentally copied to real batches (which are then returning temporaries...)

This is a step up from how it was before. It uses less concepts in a much more principled way. There are less things that can go wrong, albeit do not be mistaken: there are things that go wrong still.


Ulfgard:
merging as noone commented and tests pass.

PR #154, opened 2017-03-06
Change name of result variable in end of Socket constructor

egomeh:
The variable name 'result' was redefined when _WIN32 was set. The result
variable is now changed from 'result' to 'connectionResult'


Ulfgard:
Thanks a lot!

PR #157, opened 2017-03-08
Bugfix msvc14

egomeh:
Should fix the tests DirectSearch_HypervolumeContribution and DirectSearch_IndicatorBasedSelection on Visual Studio 2015


Ulfgard:
Hi,

I see. Sorry for my late reply this somehow dropped under my radar.

I will cherry-pick the first patch. The second patch is actually hiding a potential error. One should never select the edge points because they have undefined hypervolume. Moreover, we use them to define the hypervolume for all other points. So it is really, really bad to remove them.

PR #176, opened 2017-04-13
Include numeric header from STL in HypervolumeIndicator

egomeh:
Add the numeric header to the includes in HypervolumeIndicator

Lacking this causes compile errors on Visual C++


Ulfgard:
Good catch. Which version of VC++ are you using? I remember it has very recently compiled on MSVC14

PR #177, opened 2017-04-20
Logistic regression

Ulfgard:
I added a trainer for LogisticRegression which supports L1 and L2 regularization, with/out bias. In theory it should also support multi-class but this is untested

This required adding an extension to BFGS allowing for box constraints. With box constraints we propose steps in form of the dogleg-trust-region algorithm (just the initial step, afterwards a normal line-search is performed). This further required adding a backtracking line-search, because neither dlinmin nor wolfe could do it.

Testing confirmed that the L1-regularization can solve the PamiToy problem


Test/Algorithms/GradientDescent/LBFGS.cpp
```diff
@@ -45,10 +47,102 @@ BOOST_AUTO_TEST_CASE( BFGS_wolfe_Rosenbrock )
...
 	optimizer.lineSearch().lineSearchType()=LineSearch::WolfeCubic;
 
 	std::cout<<"Testing: "<<optimizer.name()<<" with "<<function.name()<<" and wolfe line search"<<std::endl;
 	testFunction(optimizer,function,100,100);
 }
 
+
+
+
+//with 2*n variables, variables 0...n-1 are constrained tbe in [-1,2] and the others
```
TGlas: [-1, 2] is a pretty odd range, but no problem

Test/Algorithms/Trainers/LogisticRegression.cpp
```diff
@@ -0,0 +1,157 @@
+#define BOOST_TEST_MODULE TRAINERS_LDA
```
TGlas: wrong name, should be something along the lines of TRAINERS_LOGISTICREGRESSION

Test/Algorithms/Trainers/LogisticRegression.cpp
```diff
@@ -0,0 +1,157 @@
...
+	BOOST_CHECK_EQUAL(model.numberOfParameters(),11);
+	
+	//check gradient
+	{
+		CrossEntropy loss;
+		ErrorFunction error(dataset, &model.decisionFunction(),&loss);
+		RealVector derivative;
+		error.evalDerivative(model.parameterVector(),derivative);
+		RealVector penalty = 0.1 * model.parameterVector();
+		penalty(10) = 0;
```
TGlas: Is this the bias? If so, is it possible to obtain the linear part of the function directly from the model, not indirectly through the parameter vector?

Test/Algorithms/Trainers/LogisticRegression.cpp
```diff
@@ -0,0 +1,157 @@
...
+		//create subgradient derivative
+		for(std::size_t i = 0; i != 10; ++i){
+			if(std::abs(point(i)) > 1.e-13)
+				derivative(i) += lambda1 * (point(i) > 0? 1.0 : -1.0);
+			else//otherwise choose the proper subgradient that makes the resulting gradient smallest
+				derivative(i) -= std::min(lambda1,std::abs(derivative(i))) * (derivative(i) > 0? 1.0 : -1.0);
+		}
+		
+		std::cout<<point<<std::endl;
+		std::cout<<derivative<<std::endl;
```
TGlas: Is the cout debug code?

include/shark/Algorithms/GradientDescent/LBFGS.h
```diff
@@ -69,34 +90,54 @@ class LBFGS : public AbstractLineSearchOptimizer{
...
+	///
+	/// Approximately solves the constrained optimization problem
+	/// min_p 1/2 p^TBp +g^Tp
+	/// s.t. l_i <= x_i + p_i <= u_i
+	/// This is done using a constrained dogleg approach.
+	///
+	/// first, all variables with active constraints (e.g. x_i = l_i and g_i > 0)
+	/// are fixed, i.e. p_i = 0. For the remaining variables, the unconstrained optimization
+	/// problem is solved. If the solution does not statisfy the box constraints, in the next step
+	/// the cauchy point is computed. If the cauchy point is feasible, we search the the point
```
TGlas: the the (accidental repetition)

include/shark/Algorithms/Trainers/LogisticRegression.h
```diff
@@ -0,0 +1,279 @@
...
+	typedef LinearClassifier<InputVectorType> ModelType;
+	typedef LabeledData<InputVectorType, unsigned int> DataType;
+
+	/// \brief Constructor.
+	///
+	/// \param  lambda1    value of the 1-norm regularization parameter (see class description)
+	/// \param  lambda2    value of the 2-norm regularization parameter (see class description)
+	/// \param  lbias          whether to train with bias or not
+	/// \param  accuracy  stopping criterion for the iterative solver, maximal gradient component of the objective function (see class description)
+	LogisticRegression(double lambda1 = 0, double lambda2 = 0, bool bias= 1, double accuracy = 1.e-8)
```
TGlas: please use true, not 1 :)

include/shark/Algorithms/Trainers/LogisticRegression.h
```diff
@@ -0,0 +1,279 @@
...
+
+	class L1Reformulation: public SingleObjectiveFunction{
+	public:
+		L1Reformulation(ErrorFunction* error, double lambda1, std::size_t regularizedParams)
+		: mep_error(error), m_lambda1(lambda1), m_regularizedParams(regularizedParams){
+			m_features |= CAN_PROPOSE_STARTING_POINT;
+			m_features |= HAS_FIRST_DERIVATIVE;
+			
+			std::size_t dim = numberOfVariables();
+			double unconstrained = std::numeric_limits<double>::max();
```
TGlas: In order to avoid infinite values, I prefer 1e100 for this purpose, since we can do calculations with it (i.e., adding epsilon) without having to deal with infinite values. Actual infinity is an alternative, but that may not be portable to non-IEEE complient systems.

src/Algorithms/GradientDescent/LBFGS.cpp
```diff
@@ -4,13 +4,13 @@
  * \brief       LLBFGS
  * 
  * The Limited-Memory Broyden, Fletcher, Goldfarb, Shannon (LBFGS) algorithm
- * is a quasi-Newton method for unconstrained real-valued optimization.
+ * is a quasi-step method for unconstrained real-valued optimization.
```
TGlas: something odd has happened to this comment (search&replace Newton -> step?)

Test/Algorithms/Trainers/LogisticRegression.cpp
Ulfgard: in case the test fails, i want to see that the last 5 features are reasonably close to 0. e.g. no bug but just wrong level of accuracy

PR #184, opened 2017-05-06
Replace set with vector for attribute random index generation

inglada:
Using a vector instead of a set brings a little improvement. I also use shuffling of indices instead of iterative draws and set insertion, which speeds up the generation of the index table.


Ulfgard:
Hi,

could you fix the whitespace changes? we use tabs for formatting(I hope everywhere).

Issue #211, opened 2017-12-29
Getting tree node information from Random Forests


inglada:
Hi,
I would like to access to several levels of detailed information from the leaf nodes of a RF classifier. 
Right after training, I would need to get the training samples which end up in each node of each tree. At classification time, for every test sample, I would like to retrieve a unique identifier for the leaf node where the sample falls in every tree. 
This information would be very useful in order to estimate posterior probabilities for every sample.
Could you point me in the right direction to do that? I assume that the CARTree class should be slightly modified to that, since for instance, the leaf nodes don't store a unique ID, but an index to the label array.
Thank you.

Ulfgard:
Hi,

Solving the second question first is easier. I think for this you have at least to modify the CARTree class in models/trees so that evalPattern returns nodeId instead of the label. Then, you can just rewrite eval to return the label of the node with that id (last line in evalPattern). Of course you should make evalPattern public (and maybe give it a more telling name, e.g. "findLeaf".

Using these ids you can next try to solve the other task, simply by pushing the bootstrap sample through the tree in RFTrainer and record the leaf nodes. For the information about the inner nodes, you can for every sample just traverse the path upwards after findLeaf returns and add the sample to some array of "samples of node".

For this, you need the id of the parent of a node, which we do not store. You can generate the parent information using the node routines in the RFTree class. Start at node 0 and then proceed to left and right nodes storing their parent in an array
parents[nodeId] =parentId.

I would not add additional fields to the node objects because this slows down tree traversal(cache effects). Rather add separate arrays.

Does his help? Are you implementing some well known algorithm or do you try to roll your own?

Best,
Oswin




________________________________

inglada:
Hi,
Thanks for your reply. I was not thinking about the inner nodes, but just the leaves.

If I understand what you suggest, it would be impossible to use batches and getting the nodeId would require working with individual samples. Although this is not a major problem for my first question (identifying which training samples go into each leaf), this can be very slow when doing classification of large data sets.

I thought that maybe we could add a parameter (a bool member) to the CARTree class which would tell it to return either the label or the nodeId. What do you think about that?

What I want to do is using RF as a means to build a representation of a data set. This representation would be a vector of nodeIds (one per tree in the forest). Therefore, I was thinking about making the RF return the nodeIds instead of the labels. In this case, I don't know if it makes sense to modify the RFClassifier, or rather add the trees to something like the MeanModel but which would only build the vector of nodeIds.

Thanks.

[...]

PR #215, opened 2018-01-18
CARTree returning nodeId of the leaf for a classified sample

inglada:
Hi,
I have started having a look at how to implement a solution for 
[this issue](https://github.com/Shark-ML/Shark/issues/211). The test for RFClassifier shows how to use it. I guess that the relabelTreeWithNodeIds() method should only be available for CARTrees where LabelType is unsigned int, but this is just an experiment.
Thanks for your time.


Ulfgard:
Hi, 
why did you close this? i would still like to pull parts of it (the nodeid thingy)
i think i might be able to provide you with a working code snippet for tree generation later this week, but currently i am a bit stressed :(

inglada:
Hi,

I removed the relabelling function so that you can pull from my code. I have implemented the tree generation elsewhere.

I can keep it for me, as you suggest in the issue thread, but I thought it could be useful to somebody else. What do you suggest for this kind of contribution?

Thanks.

Ulfgard:
looks really good and i think it is a good contribution complementing the tree generation routines. Very nice.

The only thing is that the indentations are not correct. we use tabs and below every if should be a level of indentation. Should i fix this or do you want to do it?

inglada:
HI,
Thanks. You can fix the indentation if it is not much work for you. Otherwise, I'll do it. What indentation style are you using (Stroustrup?)?

Ulfgard:
Done. Thank you!

We have no particular stzle that we enforce, the only rule is that everything following a { should be indented once (except class scopes and singe-line ifs). And we indent with tabs. Beyond that, there are no fixed rules.

PR #221, opened 2018-02-10
Add option for multi-threaded builds when building for Visual Studio

egomeh:
Add CMake option for multi-threaded builds on Visual Studio

When this option is enabled, Visual Studio compiles the project with all available cores yielding significantly shorter build times on some machines.


Ulfgard:
thanks!

PR #224, opened 2018-02-21
Store step population and best ever solutions in CMA-ES

egomeh:
* In addition to the best solution in a `step`, make the CMA-ES store all population values from a `step`.
* Track best ever solution on the CMA-ES.
* Some changes in style in comments and code for better consistency.
* Add interface for evolution strategy algorithms.

### DIFF
--- a/include/shark/Algorithms/AbstractEvolutionStrategy.h
+++ b/include/shark/Algorithms/AbstractEvolutionStrategy.h
@@ -45,8 +45,6 @@ namespace shark {
 /// by the user of the optimizer instead of having the optimizer handle evaluation
 /// internally.
 ///
-/// Also when init() is called as offered by the AbstractOptimizer interface, the function
-/// is required to have the CAN_PROPOSE_STARTING_POINT flag.
 template<typename IndividualType>
 class AbstractEvolutionStrategy {
 public:
--- a/include/shark/Algorithms/AbstractEvolutionStrategy.h
+++ b/include/shark/Algorithms/AbstractEvolutionStrategy.h
@@ -1,0 +1,62 @@
+//===========================================================================
+/*!
+ * 
+ *
+ * \brief       AbstractEvolutionStrategy
+ * 
+ * 
+ *
+ * \author
+ * \date        2018
+ *
+ *
+ * \par Copyright 1995-2017 Shark Development Team
+ * 
+ * <BR><HR>
+ * This file is part of Shark.
+ * <http://shark-ml.org/>
+ * 
+ * Shark is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published 
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ * 
+ * Shark is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ * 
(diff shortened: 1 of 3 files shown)


include/shark/Algorithms/DirectSearch/CMA.h
```diff
@@ -204,6 +204,20 @@ class CMA : public AbstractSingleObjectiveOptimizer<RealVector >
 		return m_numEvaluations;
 	}
 
+	///\breif Return the population form the previous step
+	std::vector<SolutionType> population() const {
```
Ulfgard: return type should be const&

include/shark/Algorithms/DirectSearch/CMA.h
```diff
@@ -204,6 +204,20 @@ class CMA : public AbstractSingleObjectiveOptimizer<RealVector >
...
+		return m_population;
+	}
+
+	///\brief Returns the best solution of the previous step
+	SolutionType best() const {
```
Ulfgard: this is already implemented in the optimizer base class

include/shark/Algorithms/DirectSearch/CMA.h
```diff
@@ -204,6 +204,20 @@ class CMA : public AbstractSingleObjectiveOptimizer<RealVector >
...
+		return m_best;
+	}
+
+	///\brief Returns the best ever discovered solution
+	SolutionType bestEver() const {
```
Ulfgard: If we want that, this should also go in the base class. But I am not sure that we want that. Often, what is considered the "all-time-best" solution would depend on a different metric, e.g. classification error of the model.

include/shark/Algorithms/AbstractEvolutionStrategy.h
```diff
@@ -0,0 +1,62 @@
...
+/// \brief An optimizer that applies an evolution stragtegy
+/// 
+/// This interface serves the purpose of allowing an evolution strategy based
+/// optimizer to supply a population set to an external evaluation mechanism
+/// as opposed to only expose the `step` function. This puts more control
+/// in the user of the optimizer as evaluation is shifted to be fullly controlled
+/// by the user of the optimizer instead of having the optimizer handle evaluation
+/// internally.
+///
+/// Also when init() is called as offered by the AbstractOptimizer interface, the function
```
Ulfgard: this is a global requirement. optimizer.init(f); requires that f can propose a starting point. See
https://github.com/Shark-ML/Shark/blob/d63962161d9e7229c804373f812015f84a121139/include/shark/Algorithms/AbstractOptimizer.h#L118

include/shark/Algorithms/DirectSearch/CMA.h
egomeh: Ah yes, I missed that.

include/shark/Algorithms/DirectSearch/CMA.h
egomeh: Alright, I'll remove it in a PR update.

include/shark/Algorithms/DirectSearch/CMA.h
egomeh: I see your point. I will remove these in a PR update. If the `best` ever is to be tracked, one might as well do this manually as the `best` is exposed per iteration.

egomeh:
I have updated the CMA-ES to inherit the `AbstractEvolutionStrategy` and make the sampling function and distribution update functions public. This would make it fairly easy to apply a custom selection mechanism.

Ulfgard:
Hi,

I will for now not accept your pull-request. we wait until your project is done and until then you can work in your own branch. i will pull in the end. I have to see everything in a broader perspective than only EA, so I might make my final design decisions slightly differently.

PR #226, opened 2018-03-31
Replacing MeanModel by Ensemble

Ulfgard:
This change replaces the MeanModel by the advanced Ensemble class. This entails that Random Forests which right now use the MeanModel are now using Ensemble

An Ensemble creates a unified response from the responses of a set of models. This is quite often just weighted averaging, however for class labels it is a bit more involved: first a weighted voting scheme is used to get a class-probability vector. then the class label with largest probability (most votes) is used.

the model is quite flexible. the user can decide whether the output is vectorial (averaged input or class probability vector) or a class label (using max-voting after averaging). If the output is a class label, this class is automatically derived from Classifier<>.

Moreover, the class allows models to be either stored explicitely (e.g. random forests) or using pointers to models stored outside. This makes the implementation slightly messy but thus also allows easy using of ensembles on heterogeneous models.


Ulfgard:
@TGlas care to review?
@inglada can you check whether any of this breaks your code-base? Serialization is probably not backwards compatible.

TGlas:
Looks good to me. Given the flexibility, it is actually not too much bloated with template magic.
I was wondering whether it is possible to create heterogeneous ensembles like this:

    Ensemble<AbstractModel<RealVector, RealVecot>*> ensemble;

and then add different model types to it, like a linear model, a neural network, and an SVM?

Ulfgard:
Yes, this is possible!

inglada:
OTB builds fine with this change. I had to update the baselines (saved models) for our unit tests, but that's all.

However, I have a PhD student who just wrote a model to use leaf node purity in the voting mechanism of RF and I guess he will have to adapt a bunch of things ... Maybe he'll ask for help.

inglada:
Hi,
Actually, there is a problem when building on MSVC:
Apparently a name collision in shark/Models/Ensemble.h:259 between the template name ModelType and a private typedef in base class ModelType. 

For clarity, the class Ensemble should be declared with something like template<class TModel, class TOutput = void>
Thanks

Ulfgard:
will fix. thanks for checking!

I will merge as soon as your phd-student says it is still working.

Btw: is informing along these lines helpful for you?

inglada:
Hi,
Thanks for your quick reply. From our point of view, you can merge once the MSVC issue is fixed.
Thanks also for keeping us informed. Now that I have switched on github notifications by e-mail, I think this is the good way of getting the info.
Cheers.

tardyb:
Hi, I am the PhD student mentioned above...

I just compiled and tested with this change and everything looks good.  
Since I have duplicated a lot of code from Random Forest (and the MeanModel) to create a model to use leaf node purity, past the disappearance of the getModel() function there are no conflict. 

However, I will have some question to improve my implementation, and also allow using the Ensemble class features to these modified RFs, which is not yet possible. 
Where can explain you what I have done?

Thanks

Ulfgard:
you can send an e-mail to my university address oswin.krause [at] di.ku.dk

PR #241, opened 2018-07-09
Rework of BatchInterfaceAdaptStruct

Ulfgard:
I adapted this beautiful Chimera of template magic and Preprocessor Shenanigans to C++11. 
I make heavy use of aggregative initialization, which restricts types to not have any explicit constructors and most importantly no base-classes.

See: https://en.cppreference.com/w/cpp/language/aggregate_initialization

This allows me to generate a lot of classes using a few preprocessor primitives, namely the ability to generate {a,b,c,d} and a=other.a; b= other.b;...

For everything to work smoothly, there must be 6 classes generate:
1. type of the batch
2. reference to an element stored in the batch(for element access)
3. const reference to...
4. proxy of a batch (to generalize he fact that LabeledData does not store batches as a whole)
5. const_proxy of...
6. shape_type: to have a meaningful generalization of shape. i.e. we also store one shape descriptor for each member of the aggregate.

further, a bit of preprocessor magic is needed to figure out "what is the third member of the struct?" this is used in MklKernel and MultiTaskKernel.

the most headache was implementing all the conversion rules. e.g. batch->proxies, proxies->batch, const_proxy->proxxy, reference->value, value->reference etc. I hope i gt everything right.

Pros of this pull-request:
1. It simplifies the code a lot, we could drop the whole dependency to boost.fusion.
2. It is nearly understandable
3. A lot of absolutely not understandable code is gone
4. for some reason our test suite is now faster....
5. It is general enough to be used by InputLabelPair and WeightedDataPair.

Cons of this pull-request: 
1.This is code i dont want anybody to review. Sorry, Tobias.
2. It only works for aggregate types

### DIFF
--- a/include/shark/Data/BatchInterfaceAdaptStruct.h
+++ b/include/shark/Data/BatchInterfaceAdaptStruct.h
@@ -262,22 +262,19 @@ static auto get_tuple_elem(T&& input, std::integral_constant<int,N>) -> decltype
 ///<code>
 ///struct DataTypeBatch{
 ///     RealMatrix A;
-///     RealVector B;
+///     Batch<T>::type B;
 ///};
 ///</code>
-///In this case the macro can be used to generate a complete specialisation of Batch<DataType>
+///In this case the macro can be used to generate a complete specialisation of Batch<DataType<T> >
 ///<code>
-///#define DataVars (RealVector, A)(double B)
-///
-///SHARK_CREATE_BATCH_INTERFACE( DataType,DataVars)
+/// template<class T>
+/// SHARK_CREATE_BATCH_INTERFACE( DataType<T>,(RealVector, A)(T, B))
 ///};
-///In this case also boost::fusion support is added to the sequence. e.g. it is
-///handled similar to any other tuple type (RealMatrix,RealVector). This is useful for MKL or Transfer
-///kernels
 ///</code>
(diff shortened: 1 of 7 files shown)


include/shark/Data/Dataset.h
```diff
@@ -119,9 +119,9 @@ class Data : public ISerializable
 	typedef Type element_type;
 	typedef typename Batch<Type>::shape_type shape_type;
 
-	typedef typename Container::BatchType batch_type;
```
TGlas: If you happen to have a refactoring tool that can replace typedef X Y; with using X = Y; then please, please apply it to the whole library.

include/shark/Data/BatchInterfaceAdaptStruct.h
```diff
@@ -292,114 +242,39 @@ public:\
...
-		return reference(batch, detail::MakeRef(i));\
-	}\
-	static const_reference get(type const& batch, std::size_t i){\
-		return const_reference(batch, detail::MakeConstRef(i));\
-	}
-
+	SHARK_CREATE_BATCH_INTERFACE_SEQ(SHARK_BATCH_MAKE_SEQUENCE(ATTRIBUTES))\
+};
+	
 #endif
```
TGlas: I don't only hate this obscure preprocessor magic, it is in a way far more intricate than templates, I also simply don't understand it. The conversions looks good though, as far as I can tell. However, this is an educated guess, at best.

PR #242, opened 2018-07-11
further refactoring of Data classes

Ulfgard:
Another step of refactoring. This time we completed the step of making Data a container of batches. 
Changes:
- SharedContainer got merged into Data
- the most important aspects for random access containers got implemented in data.
- batches() got removed, the dataset is now a container of batches
- setting a shape is done with setShape(), removed inputShape(), labelShape() etc
- globally replaced:
* numberOfBatches()->size()
* batch(i) -> operator[i]
* inputShape()->shape().input
* labelShape()->shape().label


Ulfgard:
@TGlas only Dataset.h needs to be checked.

